PR #15, opened 2019-03-07
Improvement/#7 general refactor

Wirnex:


server/src/data/index.js
```diff
@@ -0,0 +1,9 @@
+const { MeasurablesData } = require('./measurables-data');
```
OAGr: I'm not familiar with the naming convention of this being called ``data``. Where does that come from?

server/src/data/measurables-data.js
```diff
@@ -0,0 +1,156 @@
...
+      valueType,
+      expectedResolutionDate,
+      resolutionEndpoint,
+      descriptionDate,
+      descriptionEntity,
+      descriptionProperty,
+      channel
+    } = values;
+    let _auth0Id = await this.usersData.getAuth0Id(options);
+    const user = await this.usersData.auth0User(_auth0Id);
```
OAGr: While this is being done, can this be DRYd up?

server/src/data/measurables-data.js
```diff
@@ -0,0 +1,156 @@
...
+   * @param options
+   * @return {Promise<Promise<void>|Promise<WebAPICallResult>|never>}
+   */
+  async archiveMeasurable(root, values, options) {
+    const { id } = values;
+    const _auth0Id = await this.usersData.getAuth0Id(options);
+    const user = await this.usersData.auth0User(_auth0Id);
+    let measurable = await models.Measurable.findById(id);
+    if (measurable.creatorId !== user.agentId) {
+      throw new Error("User does not have permission");
```
OAGr: Error strings should also be put somewhere else / organized, I assume.

OAGr: With that, there's probably a nicer way to do authorization here. Like,
``await hasAccess(measurable.creatorId, user)``
We could wait for later though.

server/src/data/measurables-data.js
```diff
@@ -0,0 +1,156 @@
...
+  constructor() {
+    this.usersData = new UsersData();
+  }
+
+  /**
+   * @param root
+   * @param values
+   * @param options
+   * @return {Promise<*>}
+   */
```
OAGr: How useful do you find these annotations? Most seem fairly obvious to me.

server/src/data/users-data.js
```diff
@@ -0,0 +1,82 @@
+const _ = require('lodash');
```
OAGr: What do you think of using RamdaJs instead?

server/src/data/users-data.js
```diff
@@ -0,0 +1,82 @@
...
+   * @param options
+   * @return {Promise<*>}
+   */
+  async getAuth0Id(options) {
+    const res = await options.user;
+    const ok = _.get(res, 'ok');
+    const name = _.get(res, 'result.name');
+    const sub = _.get(res, 'result.sub');
+    if (!ok) throw new Error(name);
+    if (!sub) throw new Error("No User Id");
```
OAGr: Should we have types of errors soon? (Could be a separate story)

server/src/data/users-data.js
```diff
@@ -0,0 +1,82 @@
...
+   * @param options
+   * @return {Promise<Model>}
+   */
+  async editUser(root, values, options) {
+    const { id, name } = values;
+    let _auth0Id = await this.getAuth0Id(options);
+    let user = await models.User.findById(id);
+    if (user && user.auth0Id === _auth0Id) {
+      user.update({ name });
+    }
```
OAGr: If this is not the case, I imagine there should be an error.

server/src/data/users-data.js
```diff
@@ -0,0 +1,82 @@
...
+  async getUser(ops, values, options){
+    const { id, auth0Id } = values;
+
+    let _auth0Id = await this.getAuth0Id(options);
+    const _auth0User = await this.auth0User(_auth0Id);
+
+    let user;
+    if (_auth0Id && !_auth0User) {
+      try {
+        user = await models.User.create({ auth0Id: _auth0Id, name: "" });
```
OAGr: This is obviously quite hacky. I suggest thinking about a better way of doing this.

OAGr: Maybe doing a bit of investigation into auth0

server/src/models/user.js
```diff
@@ -1,37 +1,36 @@
...
-      defaultValue: Sequelize.UUIDV4,
-      allowNull: false,
-    },
-    name: {
-      type: DataTypes.STRING,
-      allowNull: false,
+      id: {
+        type: DataTypes.UUID(),
+        primaryKey: true,
+        defaultValue: Sequelize.UUIDV4,
```
OAGr: Now that I look at it, this seems like a mistake. It's weird to mix ``Sequelize`` with ``DataTypes``

OAGr:
Good work! I've left some comments, let's discuss those a bit, then happy to merge.

server/src/data/measurables-data.js
Wirnex: they help with code completion http://usejsdoc.org
also we can use typescript d.ts files to do better IDE code completion

server/src/data/users-data.js
Wirnex: Never used it before, it seems to be more functional style than lodash. Do you prefer RamdaJs?

server/src/models/user.js
Wirnex: `type: DataTypes.UUID,
defaultValue: DataTypes.UUIDV1` probably? http://docs.sequelizejs.com/variable/index.html

server/src/data/index.js
Wirnex: I don't have proof links but in couple projects I'm working on we have a middleware, data and model layers. Middleware handles requests and uses data layer to do CRUD operations. Data layer is an abstract layer, it interacts with models. Data isn't aware of requests, responses and database stuff.

server/src/data/measurables-data.js
Wirnex: let's refactor and write tests for that in a separate task

server/src/data/index.js
OAGr: Interesting. I've heard similar things be called Services or similar. Data is fine for now; the organization is more important than the naming.

server/src/data/measurables-data.js
OAGr: Sounds good

server/src/data/measurables-data.js
OAGr: Sounds good. Happy to discuss/move to typescript. I do prefer reasonML though, so am hesitant to change if it would be a lot of work, and we may change again in 9-20 months.

server/src/data/users-data.js
OAGr: Yea. I haven't used it a whole lot, but want to move more in that direction. I prefer to stay more functional where possible.

server/src/models/user.js
OAGr: Sounds good

OAGr:
I made these two separate issues for the future:
https://github.com/OAGr/foretold/issues/20
https://github.com/OAGr/foretold/issues/19

PR #18, opened 2019-03-08
Feature/#17 docker compose project

Wirnex:
We need this for quick deployment of our dev environment. Also we can run production containers in Heroku and other platforms like Kubernetes.

### DIFF
--- a/server/docker-compose.yml
+++ b/server/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - PORT=80
       - NODE_PATH=/opt/app/lib
     command: >
-      node_modules/.bin/nodemon lib/src/index.js --exec "node -r dotenv/config -r babel-register"
+      node_modules/.bin/nodemon src/index.js --exec "node -r dotenv/config -r babel-register"
 
   database:
     image: postgres:11
--- a/server/docker-compose.yml
+++ b/server/docker-compose.yml
@@ -4,12 +4,11 @@ services:
 
   app:
     build: ./
-    ports:
-      - 28750:80
     volumes:
       - ./src:/opt/app/lib/src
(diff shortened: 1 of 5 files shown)


Wirnex:
We need this for quick deployment of our dev environment. Also we can run production containers in Heroku and other platforms like Kubernetes.

server/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+  app:
+    build: ./
+    volumes:
+      - ./src:/opt/app/src
+      - ./migrations:/opt/app/migrations
+      - ./seeders:/opt/app/seeders
+      - ./config:/opt/app/config
+    environment:
+      - AUTH0_SECRET=<change-me>
+      - SLACK_WEBHOOK_URL=<change-me>
```
OAGr: Is it possible to use local ENV variables instead of manual editing of these variables?

Wirnex: what do you mean local ENV?
these variables will appear in process.env like process.env.AUTH0_SECRET @OAGr

OAGr: why do they say <change-me>? Is that supposed to be modified directly to the file?

OAGr: I meant how the local env variables that the user who is running this would use, in order to get the variables to be understood by Docker.

Wirnex: if you want some environment var to be defined in a docker container you should set it in .yml file. But may be you want something like this `environment:
    - HAPI_DB_CONNECT=${NODE_DB_CONNECT}` ? (haven't check it) In this case local env NODE_DB_CONNECT value should be used https://stackoverflow.com/questions/30494050/how-do-i-pass-environment-variables-to-docker-containers

PR #27, opened 2019-03-11
Improvement/#13 mongo like ids

Wirnex:
need to run migrations


OAGr:
Closing for now, we can bring this back later (to save heroku costs)

PR #39, opened 2019-03-14
Improvement/#4 client servings

GalochkinKirov:


OAGr:
Nice work so far. 
Here's the corresponding url:
https://prediction-backend-stagi-pr-39.herokuapp.com

I've noticed that corresponding urls are broken. So /login doesn't work.

client/package.json
```diff
@@ -56,10 +57,9 @@
     "react-icons-kit": "^1.2.1",
     "reason-apollo": "^0.13.0",
     "reason-react": ">=0.4.0",
+    "static-server": "^2.2.1",
     "subscriptions-transport-ws": "^0.9.14",
-    "victory": "^32.0.0"
-  },
-  "devDependencies": {
+    "victory": "^32.0.0",
```
OAGr: Why is this a devDevependency?

server/src/index.js
```diff
@@ -52,10 +53,15 @@ const server = new ApolloServer({
 const app = express();
 const cors = require("cors");
 app.use(cors());
+app.get(/^((?!graphql).)*$/,
```
OAGr: Can you add a comment to explain this a bit?

client/package.json
Wirnex: It's not, it's the **dependencies**

OAGr: Oh sorry, I think I misread the commit diff.

server/src/index.js
GalochkinKirov: Yes, sure, did

OAGr:
Nice work on this!

PR #44, opened 2019-03-18
Improvement/#8 channels

GalochkinKirov:


server/src/models/agent-channel.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @param sequelize
+ * @param DataTypes
+ * @return {*}
+ */
+module.exports = (sequelize) => {
+  const AgentChannel = sequelize.define('AgentsChannels', {}, {
+    timestamps: false
```
OAGr: Why no timestamps on these? A createdAt may be nice.

server/migrations/20190318001100-addUuidExt.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  up: async function (queryInterface) {
+    return queryInterface.sequelize.query(`CREATE EXTENSION "uuid-ossp"`);
```
OAGr: Why this instead of the previous way? If we are using ``uuid_generate_v4``, should we have another story later to convert previous tables to use this as well?

server/migrations/20190318001110-channelsTable.js
```diff
@@ -0,0 +1,50 @@
+module.exports = {
+  up: async function (queryInterface, Sequelize) {
+    await queryInterface.createTable('Channels', {
+      id: {
+        type: Sequelize.UUID,
+        primaryKey: true,
+        defaultValue: Sequelize.fn('uuid_generate_v4'),
```
OAGr: Why this instead of the previous way (which I think was done in the model layer)?

Wirnex: isn't it better to do the generation on the database side? you can use it even in raw queries
then we can migrate other tables to this default value too

server/migrations/20190318001115-addDefChannel.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = {
+  up: async function () {
+    const user = await models.User.findOne({
+      where: { name: 'Service Account' },
+    });
+    await models.Channel.findOrCreate({
+      where: { name: 'unlisted' },
+      defaults: {
+        name: 'unlisted',
+        agentId: user.agentId,
```
OAGr: From above, it doesn't look like channels have agentIds.

server/migrations/20190318001121-agentsChannels.js
```diff
@@ -0,0 +1,26 @@
...
+      },
+      channelId: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: 'Channels',
+          key: 'id',
+        }
+      },
+    });
```
OAGr: Can this also have a createdAt attribute or similar?

server/migrations/20190318001122-addChannelId.js
```diff
@@ -0,0 +1,34 @@
...
+    `);
+
+    await queryInterface.sequelize.query(`
+        ALTER TABLE "Series" ADD COLUMN "channelId" uuid;
+        UPDATE "Series" SET "channelId" = '${channel.id}';
+        ALTER TABLE "Series" ALTER COLUMN "channelId" SET NOT NULL;
+
+        ALTER TABLE "Series" ADD CONSTRAINT "Series_channelId_fkey"
+        FOREIGN KEY ("channelId") REFERENCES "Channels" ("id")
+        on update cascade on delete set null;
```
OAGr: nice work with all this SQL

server/src/models/channel.js
```diff
@@ -0,0 +1,45 @@
...
+      default: true,
+    },
+  });
+
+
+  Channel.associate = function (models) {
+    Channel.Creator = Channel.belongsTo(models.Agent, {
+      foreignKey: 'creatorId',
+      as: 'creator'
+    });
```
OAGr: Can you also add ``Chanel.Agents`` or similar, that lists all members, via ``AgentChannel``, then do the same for an agent (``Agent.Channels``)

OAGr: This could also be part of a different story, if there's another specific one in mind.

server/migrations/20190318001110-channelsTable.js
GalochkinKirov: it's like single way of truth

OAGr: I like that, makes sense, just checking.

GalochkinKirov:
We have added tests for new channel model and its associations,
to run it type the command:
`npm run babel-node ./src/__tests__/manuals/channels.js`

OAGr:
I'm not super happy with these specific tests, but given that there's further work on this don't want to hold it back. Will merge, and we can revisit later.

PR #49, opened 2019-03-21
Feature/#29: Channels: Simple Creation (Queries & Mutations)

GalochkinKirov:


server/src/data/agents-channels-data.js
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * @tested
+   * @param {string} channelId
+   * @param {string} agentId
+   * @returns {Promise<Model>}
+   */
+  async createOne(channelId, agentId) {
+    const input = { channelId, agentId };
+    await this.validate(input);
```
OAGr: I don't like the ``validate`` name. Maybe something like ``agentHasChannelAccess`` or something else that's more semantic?

OAGr:
Overall this looks good to me. Not sure about some of the details, especially around the testing, but we can figure those out more later. Need to merge this sooner rather than later for some UI work, so will merge.

PR #52, opened 2019-03-26
Improvement/#30 shield middlewares

GalochkinKirov:


server/src/authorizers/index.js
```diff
@@ -0,0 +1,40 @@
+const { shield } = require('graphql-shield');
+
+const { isChannelAllowed } = require('./channels');
+
+const permissions = shield({
+  Query: {
+    // user: not(isAuthenticated),
```
OAGr: why are all of these commented out?

server/src/middlewares/measurements.spec.js
```diff
@@ -0,0 +1,66 @@
...
+
+  describe('measurement ', () => {
+    const measurementObj = { measurableId: 'measurableId1' };
+    beforeAll(() => {
+      jest.spyOn(data.measurementData, 'getOne').mockReturnValue(
+        Promise.resolve(measurementObj),
+      );
+    });
+
+    it('A', () => {
```
OAGr: Can you give a description? "A" doesn't mean anything to me.

server/src/middlewares/series.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @tested
+ * @param {object | null} root
+ * @param {object} args
+ * @param {object} context
+ * @param {object} info
+ * @return {Promise<void>}
+ */
+async function serie(root, args, context, info) {
```
OAGr: what is "serie"? The singular of "series"? If so, it could be "seriesSingle" or similar. Series is kind of a strange word, in that its singular and plural are the same.

server/src/models/__mocks__/index.js
```diff
@@ -26,9 +23,42 @@ const User = {
   findOrCreate: jest.fn(() => Promise.resolve(true)),
 };
 
+const Measurable = {
+  findOne: jest.fn(() => Promise.resolve(true)),
```
OAGr: This is not very dry. Can you at either make one fn for ``jest.fn(() => Promise.resolve(true))``, or go further and have a function to generate these?

server/src/schema.js
```diff
@@ -32,36 +37,40 @@ const schema = new graphql.GraphQLSchema({
...
+
       measurables: {
         type: new graphql.GraphQLNonNull(new graphql.GraphQLList(types.measurableType)),
         args: {
           offset: { type: graphql.GraphQLInt },
           limit: { type: graphql.GraphQLInt },
           creatorId: { type: graphql.GraphQLString },
-          channelId: { type: graphql.GraphQLString },
           seriesId: { type: graphql.GraphQLString },
+          channelId: { type: graphql.GraphQLNonNull(graphql.GraphQLString) },
```
OAGr: Can this stay nullable? It would be nice if a user could call it without channelId, and then it figures out which channels that user has access to, then gathers all the measurables.

server/src/resolvers/measurements.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    jest.clearAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  describe('all', () => {
+    const root = {};
+    const args = {};
+    const context = {};
+    const info = {};
+    it('all', () => {
```
OAGr: These should have descriptions. Like,
``it('calls correct model and produces correct result`)``

server/src/authorizers/index.js
GalochkinKirov: Fixed.

OAGr:
Other comments: 
- http://localhost:4000/graphql should be accessible. This PR requires a token, which means that that breaks. I think it should be possible to at least see the schema without a token.
- After making server changes (but before making a PR), please go to the client and run ``yarn run update-gql``. This will update the graphql schema on the frontend, which will also be useful for making it obvious how the schema changed.

server/src/authorizers/channels.spec.js
```diff
@@ -0,0 +1,49 @@
+const channels = require('./channels');
+
+describe('tests channel authorizers', () => {
```
OAGr: This is better. 
A few things though:
1) Don't use the word "tests" in tests. It's obvious it's a test. Instead, just have things like ``Channel Authorizers``
2) Don't use the word "should" in tests. Just phrase them like, "returns true if channel is public"
3) Use other describe blocks for context. For instance, ``describe(`'when the channel is public`)`` and have tests that describe states where the channel is public.

server/src/data/channels-data.js
```diff
@@ -71,6 +71,68 @@ class ChannelsData {
...
 
+  /**
+   * @public
+   * @param {object} options
+   * @param {number} [options.offset]
+   * @param {number} [options.limit]
+   * @param {object} [options.restrictions]
+   * @return {Promise<*|Array<Model>>}
+   */
+  async getAll(options = {}) {
```
OAGr: given that this is already in data, can we replace the ``options`` input with a specific hash of keys, like, 
``getAll({offset=0, limit=10})``. Not sure exactly what getRestrictionsSync needs though.

server/src/authorizers/permissions.js
```diff
@@ -0,0 +1,20 @@
...
+
+function getPermissions() {
+  return shield({
+    Query: {
+      measurement: isChannelAllowed,
+      measurements: isChannelAllowed,
+      measurable: isChannelAllowed,
+      measurables: isChannelAllowed,
+      series: isChannelAllowed,
+      seriesCollection: isChannelAllowed,
```
OAGr: This feels like the wrong way to do things. I'd prefer it if a user could query for all measurables/measurements, but only get back ones that belong to all the channels they have access to. They shouldn't need to declare one specific channel upfront.

server/src/authorizers/channels.spec.js
GalochkinKirov: Good point, thanks.

server/src/models/__mocks__/index.js
GalochkinKirov: No, it is not a principle of DRY when you use auto-generation function in THIS case. We must call jest.fn in each case where we want to get Jest Mock Object. Each of these objects has own state.

Wirnex: It should be fine with mocks this way because it's clear what a mock returns. And we should keep them as different objects with it's own context (calls count, arguments passed).

server/src/data/channels-data.js
GalochkinKirov: Yes, you can write code this way but believe it is not safe and not flexible. Just image that function gets "string" as param.

server/src/middlewares/index.js
```diff
@@ -0,0 +1,61 @@
...
+const { measurable } = require('./measurables');
+const { measurement } = require('./measurements');
+const { serie } = require('./series');
+
+const middlewares = {
+  Query: {
+
+    // @tested
+    measurements: async (resolve, root, args, context, info) => {
+      await measurable(root, args, context, info);
```
OAGr: This seems a bit awkward to me (though kudos for getting it to this stage). Do we really need to pass all 4 parameters into each function?

GalochkinKirov: Yes, we do.

GalochkinKirov: If one function will take "root, info", another "args, context" it would be harder to remember signature.

server/src/middlewares/measurements.spec.js
GalochkinKirov: Fixed.

server/src/resolvers/measurements.spec.js
GalochkinKirov: Fixed.

server/src/models/__mocks__/index.js
OAGr: Ok, that works.

server/src/middlewares/series.js
GalochkinKirov: Thanks, will fix.

server/src/schema.js
GalochkinKirov: Ok, will fix.

server/src/middlewares/series.js
GalochkinKirov: Fixed.

GalochkinKirov:
Temporary closing this PR.

server/src/authorizers/permissions.js
GalochkinKirov: Fixed.

server/src/schema.js
GalochkinKirov: Fixed.

server/src/middlewares/index.js
OAGr: Ok, no need to take further action now.

server/src/middlewares/index.js
```diff
@@ -0,0 +1,9 @@
+const middlewares = {
+  Query: {
+    //
+  },
```
OAGr: Are we not using the middlewares anymore? If that's the case, is there any reason to keep them around?  We could delete, and we have them in git history if needed later.

server/src/data/measurables-data.spec.js
```diff
@@ -0,0 +1,71 @@
...
+          "limit": 2,
+          "offset": 1,
+          "order": [["createdAt", "DESC"]],
+          "where": {
+            "$and": [
+              { "channelId": { "$in": "channelIdsLiteral" } },
+              { "channelId": "3" }
+            ],
+            "creatorId": { "undefined": "5" },
+            "seriesId": { "undefined": "4" },
```
OAGr: Why are these undefined?

GalochkinKirov: Fixed.

server/src/middlewares/index.js
GalochkinKirov: Thanks, fixed.

PR #53, opened 2019-03-28
Feature/#11 rls

GalochkinKirov:


server/migrations/20190325001122-rolesCol.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  up: async function (queryInterface) {
+    await queryInterface.sequelize.query(`
+        ALTER TABLE "AgentsChannels" ADD COLUMN "roles" VARCHAR(8)[];
+        UPDATE "AgentsChannels" SET "roles" = ARRAY['viewer', 'admin'];
```
OAGr: Why didn't you use the regular way of modifying tables here? Also, is this a required attribute? (And if so, make sure defaults are set, probably to viewer)

GalochkinKirov: Thanks, fixed.

server/src/middlewares/index.js
```diff
@@ -0,0 +1,39 @@
...
+      await channel(root, args, context, info);
+      await agentsChannels(root, args, context, info);
+      return await resolve(root, args, context, info);
+    },
+  }
+};
+
+module.exports = {
+  middlewares,
+};
```
OAGr: How are the middlewares used in this story? The connection isn't clear to me.

GalochkinKirov: Middlewares prepare context of users requests. Authorizers use the context. That is all.

server/src/data/channels-data.js
```diff
@@ -22,14 +22,18 @@ class ChannelsData extends DataBase {
...
+      return Promise.reject(new Error('Channel exists.'));
     }
+    channel = await models.Channel.create({
+      ...input,
+      creatorId: user.agentId,
+    });
+    await this.agentsChannelsData.createOne(
+      channel.id,
+      user.agentId,
+      models.AgentsChannels.ROLE.ADMIN,
```
OAGr: Can this be done in an ``afterCreate`` callback in the Channel model layer? That seems like where I would more expect it.

server/src/middlewares/index.js
OAGr: Ah, thanks.

server/src/data/channels-data.js
GalochkinKirov: No, according to our strategy models do not know logic of application work. They only represents data from DB.

PR #55, opened 2019-03-29
Feature/#10 perm

GalochkinKirov:


OAGr:
Good work, that was quite quick.

PR #57, opened 2019-04-01
Improment/#54

GalochkinKirov:


OAGr:
I merged this too soon. It passed the main tests, so I thought it was fine, but it wasn't. 

When I run this locally the migration fails. I've tried dropping/recreating my local table and it's still an issue.

I noticed this failed on Heroku. I assumed that was a different issue, but now think it could well be this. 

image

PR #66, opened 2019-04-04
Improvement/#40

GalochkinKirov:


OAGr:
Good work with this.
Some other things I think should be part of this:
- The new state should be accessible on query requests of measurables.
- The old state (the one that was a string, straight from the database) should not be accessible via the api.

GalochkinKirov:
> Good work with this.
> Some other things I think should be part of this:
> 
> * The new state should be accessible on query requests of measurables.
> * The old state (the one that was a string, straight from the database) should not be accessible via the api.

It is not a New State attribute, it is the representation of state string. State string is the same, but there is new possible value "JUDGEMENT_PENDING". This is why "state" accessible, but number representation of state no.

Or I missed something in your questions.

OAGr:
Hm... I'll try it out and see what you mean.

OAGr:
Yea; right now ``state`` on ``measurable`` returns a string!. It should instead return a ``measurableState!``

GalochkinKirov:
Ok, got it. Going to fix.

PR #73, opened 2019-04-04
Feature/#63 states of measurables

GalochkinKirov:


server/src/async/events.js
```diff
@@ -0,0 +1,8 @@
+const EVENTS = {
+  MEASURABLES_STATES: 'MEASURABLES_STATES',
```
OAGr: If this is an EVENT, can we change it to something like, ``RUN_MEASURABLE_STATE_TRANSITIONS``

server/src/async/listeners.js
```diff
@@ -0,0 +1,32 @@
+const emitter = require('./emitter');
+const events = require('./events');
+
+const triggers = require('./triggers');
+
+function measurablesStates() {
+  try {
+    console.log('Job::measurablesStates');
```
OAGr: please rename this as well.

server/src/async/listeners.js
```diff
@@ -0,0 +1,32 @@
+const emitter = require('./emitter');
+const events = require('./events');
+
+const triggers = require('./triggers');
+
+function measurablesStates() {
```
OAGr: rename to be more like an action

server/src/async/index.js
```diff
@@ -0,0 +1,22 @@
+const cron = require('node-schedule');
+
+const emitter = require('./emitter');
+const events = require('./events');
+
+function runJobs() {
+  console.log('Jobs are in a queue.');
+
+  const EVERY_DAY_AT_MIDNIGHT = '0 0 * * *';
```
OAGr: Can we run this every hour instead? Also, is it possible to trigger remotely for testing?

server/src/models-abstract/measurable-model.js
```diff
@@ -25,6 +29,26 @@ class MeasurableModel extends ModelPostgres {
     );
   }
 
+  /**
+   * @return {Promise<boolean>}
+   */
+  setJudgementPending() {
+    return this.model.update({
```
OAGr: I kind of hate to say this, but can we move this logic up out of the db layer and into the model layer or similar? So the model layer has a method for ``shouldupdateStateToJudgementPending``, and another for ``updateStateToJudgementPending``. That way we could later have hooks here to add notifications, and other things. For instance, the ``resulution.js`` bot should run here.

server/src/models/measurable.js
```diff
@@ -35,24 +37,12 @@ module.exports = (sequelize, DataTypes) => {
...
     },
-    resolutionEndpoint: {
-      type: DataTypes.TEXT,
-      allowNull: true,
-    },
-    hasResolutionEndpointResolved: {
-      type: DataTypes.BOOLEAN,
-      allowNull: false,
-      defaultValue: false,
-    },
```
OAGr: Why did you move these down?

server/src/async/events.js
GalochkinKirov: Event it is not a command.

OAGr: Even so. Even ``MEASURABLE_STATE_TRANSITIONS`` would be more specific.

GalochkinKirov: Thanks, ok.

PR #96, opened 2019-04-09
Feature/#21

GalochkinKirov:
Please, consider the previous PR
https://github.com/OAGr/foretold/pull/95
After that I will update this PR.


OAGr:
After some thought, I think I'm quite happy with this. Will need to spend some time with it once merged to see how exactly it works with the client.

PR #98, opened 2019-04-11
Feature/#75 2

GalochkinKirov:


server/src/models-abstract/series-model.spec.js
```diff
@@ -0,0 +1,7 @@
+const { SeriesModel } = require('./series-model');
+
+describe('Series Model', () => {
+  it('class should be a constructor', () => {
+    expect(SeriesModel).toBeInstanceOf(Function);
```
OAGr: How useful are these tests to you? I'm not sure they're worth the effort to write them, I'm not sure how much they would catch.

GalochkinKirov: They are very useful cuz they test dependencies (import/export).

PR #108, opened 2019-04-16
Feature/#100

GalochkinKirov:


server/src/authorizers/permissions.js
```diff
@@ -34,8 +34,8 @@ const rules = {
     measurements: allow,
     measurable: allow,
     measurables: allow,
-    bot: allow,
-    bots: allow,
+    bot: isAuthenticated,
+    bots: isAuthenticated,
```
OAGr: Why this restriction?

server/src/data/bots-data.js
```diff
@@ -20,6 +20,33 @@ class BotsData extends DataBase {
...
+   * @param {string} [options.userId]
+   * @param {number} [options.offset]
+   * @param {number} [options.limit]
+   * @return {Promise<*|Array<Model>>}
+   */
+  async getAll(options) {
+    const pagination = {
+      offset: options.offset,
+      limit: options.limit,
+    };
```
OAGr: Do you think it would be reasonable to use the edges/nodes pagination instead?

server/src/types/bots.js
```diff
@@ -15,6 +15,11 @@ const bot = new graphql.GraphQLObjectType({
     agentId: { type: graphql.GraphQLString },
     userId: { type: graphql.GraphQLString },
 
+    jwt: {
+      type: graphql.GraphQLNonNull(graphql.GraphQLString),
+      resolve: require('../resolvers/authentications').getJwtByAgentId,
+    },
```
OAGr: It seems like the jwt should only be available to the owner. I'm not sure how to do this in permissions, but it seems important to figure out. (row-level permissions)

server/src/authorizers/permissions.js
GalochkinKirov: Because User can get only its own bots. This is why User should be authenticated.

server/src/data/bots-data.js
GalochkinKirov: Yes, we should use only one way of pagination. Should I repair this right in the PR?

server/src/types/bots.js
GalochkinKirov: Here we can use low-level restrictions (for instance users can see only its bots) or use "shield" permissions.

server/src/data/bots-data.js
OAGr: Yea, that sounds good to me.

server/src/types/bots.js
OAGr: I think users should be able to see all bots (the names/Ids/descriptions), but not the jwt token. Any idea on the best way of doing that? Not sure if shield helps with that.

GalochkinKirov: Fixed.

server/src/authorizers/permissions.js
GalochkinKirov: Fixed.

server/src/data/bots-data.js
GalochkinKirov: Fixed.

server/src/types/bots.js
GalochkinKirov: Fixed.

GalochkinKirov:
Fixed.

server/src/authentication.js
```diff
@@ -107,6 +107,19 @@ async function getJwtByAuth0Jwt(token) {
...
 }
 
+/**
+ * @param {string} agentId
+ * @return {Promise<string>}
+ */
+async function getJwtByAgentId(agentId) {
+  try {
+    const payload = {};
+    return encondeJWT(payload, agentId);
```
OAGr: I just realized this should be ``encodeJWT``

PR #111, opened 2019-04-16
Feature/#101

GalochkinKirov:


server/src/middlewares/channel-memberships.js
```diff
@@ -18,11 +18,11 @@ async function channelMemberships(root, args, context, info) {
     || _.get(context, 'channel.id');
   const agentId = _.get(context, 'agent.id');
 
-  const id = { agentId, channelId };
-  console.log('\x1b[36m ---> \x1b[0m Middleware (channelMemberships)', id);
+  const compountId = { agentId, channelId };
```
OAGr: compount? Can you use a different word? Maybe ``combinedId`` or similar?

PR #147, opened 2019-04-22
Improvements/#143

GalochkinKirov:


OAGr:
This seems good for ``first``/ ``after``, but seems broken for ``before`` / ``last``

image

OAGr:
Seems to work now, will do the corresponding client work and then merge.

Issue #130, opened 2019-04-18
Value (floatCdf and floatPoint) should have simple validations


OAGr:
- Only one of the two can be entered.
- If a floatCdf is entered, then the xs and ys must have the same length.
- If a floatCdf is entered, the length of the xs must be 1000 or less.

These errors should return errors with descriptions of the errors.

PR #148, opened 2019-04-22
Feature/#130

GalochkinKirov:
https://github.com/OAGr/foretold/issues/130

### DIFF
--- a/server/src/middlewares/index.js
+++ b/server/src/middlewares/index.js
@@ -64,6 +64,7 @@ const middlewares = {
     },
 
     measurementCreate: async (resolve, root, args, context, info) => {
+      await measurementValueValidation(root, args, context, info);
       await measurable(root, args, context, info);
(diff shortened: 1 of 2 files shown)


OAGr:
Looks pretty good. Some comments:

Can we change the errors to: 
```
const ERR_1 = 'You have entered both floatCdf and floatPoint values. You can only submit one type of value. ';
const ERR_2 = 'Xs and Ys should be the same size.';
const ERR_3 = `Xs of length {${X_ENTERED}} exceeds maximum of length ${MAX_XS}.`;
```

Also, add an error,
```
const ERR_4 = "You must submit one kind of value." (If no values are added)

As to the ``measurementValueValidation`` function, I think it can be cleaned up & dried a bit. For one, 
I like the ``!!floatCdf`` convention instead of ``floatCdf !== undefined``

GalochkinKirov:
Fixed.

PR #149, opened 2019-04-22
wip Fixes channelMembershipDelete permissions

GalochkinKirov:


OAGr:
Good to see. 

I don't quite understand though. Can we come up with more precise names for the params? 
Maybe:
``isObjectAdmin`` -> ``isMembershipOfAnAdmin``
``isSubjectAsObject`` -> ``membershipBelongsToCurrentUser``
``isMoreThenOneAdmin`` -> ``channelHasMultipleAdmins``

OAGr:
I think one thing that's confusing here is that ``is`` represents both the current user and sometimes the channel. It may be nice to eventually refactor to be more precise about all of these, like, ``currentUserIsAdmin`` and ``currentUserIsAuthenticated``. Maybe use "requestingUser" instead of "currentUser". Thoughts? We can also do in a separate story.

PR #182, opened 2019-04-29
Feature/#173

GalochkinKirov:


cdf/cdf.js
```diff
@@ -0,0 +1,143 @@
...
+   */
+  constructor(xs, ys) {
+    this.xs = xs;
+    this.ys = ys;
+
+    if (!this.validateSize(this.xs, this.ys)) {
+      const corrected = this.convertToNewLength(this.xs.length);
+      this.xs = corrected.xs;
+      this.ys = corrected.ys;
+    }
```
OAGr: I think it should throw an error instead because we can't be sure this is correct.

cdf/cdf.js
```diff
@@ -0,0 +1,143 @@
...
+      if (a.xs > b.xs) return 1;
+      if (a.xs < b.xs) return -1;
+      return 0;
+    });
+
+    const XS = sorted.map(v => v.xs);
+    const YS = sorted.map(v => v.ys);
+
+    return { xs: XS, ys: YS };
+  }
```
OAGr: For these to be valid CDFs, when Xs are sorted, Ys should also be sorted (only being equal or going up). Can you write another check to make sure that is the case?

cdf/cdf.js
```diff
@@ -0,0 +1,143 @@
...
+   * This should do the same thing as `findY`, but for Y.
+   * @param y
+   * @return {*}
+   */
+  findX(y) {
+    let firstHigherIndex = this.ys.findIndex(Y => Y > y);
+    let lowerOrEqualIndex = firstHigherIndex - 1;
+    let needsInterpolation = this.ys[lowerOrEqualIndex] !== y;
+    if (needsInterpolation) {
+      // @todo: should we turn axes?
```
OAGr: Does this still need to be done? I assume it's a technical question, rather than a feature question (one for me)

cdf/cdf.js
```diff
@@ -0,0 +1,143 @@
...
+   * @return {number[]}
+   */
+  sample(size) {
+    return Array(size).map(() => this.sampleSingle());
+  }
+
+  //FUTURE TODO: Don't do this one in this story, but it would be nice to do later.
+  mean() {
+
+  }
```
OAGr: Do you have thoughts on a good way to do this? It's really ``meanX``.

cdf/functions.js
```diff
@@ -0,0 +1,95 @@
...
+ * @param {number} max
+ * @param {number} n
+ * @return {number[]}
+ */
+function range(min, max, n) {
+  if (n <= 0) throw new RangeError('n is less then zero');
+  if (n === Infinity) throw new RangeError('n is Infinity');
+  if (n === 0) return [];
+  if (n === 1) return [min];
+  if (n === 2) return [min, max];
```
OAGr: nice work laying out all the conditions.

cdf/functions.js
```diff
@@ -0,0 +1,95 @@
...
+
+  let item = min;
+  do {
+    result.push(item);
+    item += interval;
+  } while (item <= max);
+
+  // corrects results
+  if ((n + 1) - result.length === 1) {
+    result.push(b);
```
OAGr: what is b? I don't think I understand this bit.

OAGr:
Nice work with this. I made some comments, but those could mostly be a separate issue (except for the ``b`` one).

cdf/cdf.js
GalochkinKirov: Yes, sure, going to add tests later in another issue

cdf/cdf.js
GalochkinKirov: Fixed.

cdf/cdf.js
GalochkinKirov: I am not sure that got you properly here

cdf/cdf.js
GalochkinKirov: Not a big issue I think to get "mean" of all XS.

cdf/functions.js
GalochkinKirov: Thanks, was waiting time for testing

GalochkinKirov:
A huge thanks, fixed some lines. Going to write test in the next issue PR.

cdf/cdf.js
OAGr: I mean, can you explain why the ``todo`` is there? Is that a question for me, or something you intend to fix in this PR or next?

cdf/cdf.js
OAGr: Not exactly. The Ys matter, it's a special type of mean. I could do this one later.

Issue #175, opened 2019-04-25
Aggregation bot should use CDF library to combine distributions


OAGr:
It should combine the CDFs of the most recent competitive entry of each agent competitor, then post that as it's aggregation.

PR #185, opened 2019-04-30
Feature/#175

GalochkinKirov:
https://github.com/OAGr/foretold/issues/175

### DIFF
--- a/bots/bots/aggregation-bot.js
+++ b/bots/bots/aggregation-bot.js
@@ -56,6 +56,7 @@ class AggregationBot {
 
     if (cdfs.length === 0) return null;
 
+    // @todo: What's about the limit?
     const combined = new CdfCombination(cdfs).combine(10);
 
     return {
--- a/bots/bots/aggregation-bot.js
+++ b/bots/bots/aggregation-bot.js
@@ -1,5 +1,6 @@
 const _ = require('lodash');
 const { Cdf } = require('cdf/cdf');
+const { CdfCombination } = require('cdf/cdf-combination');
 
 const { API } = require('../api');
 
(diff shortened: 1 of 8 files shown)


bots/bots/aggregation-bot.js
```diff
@@ -13,29 +17,53 @@ class AggregationBot {
...
+    }).map((measurement) => {
+      const xs = _.get(measurement, 'value.floatCdf.xs');
+      const ys = _.get(measurement, 'value.floatCdf.ys');
+      return new Cdf(xs, ys);
+    });
+
+    if (cdfs.length === 0) return null;
+
+    // @todo: What's about the limit?
+    const combined = new CdfCombination(cdfs).combine(10);
```
OAGr: The sample count should be 1000, but should probably also be a constant on the top of the file or config or similar

bots/bots/aggregation-bot.js
```diff
@@ -13,29 +17,53 @@ class AggregationBot {
...
+      if (!aggregated) continue;
+
+      await this.api.measurementCreateAggregation({ ...id, ...aggregated });
     }
 
     return true;
   }
 
   /**
+   * @todo: What to do with "floatPoint" spot?
```
OAGr: points should never be aggregated. Points can only be final values, not competitive measuremenets.

bots/bots/aggregation-bot.js
GalochkinKirov: Ok, will update

bots/bots/aggregation-bot.js
GalochkinKirov: ok

bots/bots/aggregation-bot.js
```diff
@@ -13,29 +17,53 @@ class AggregationBot {
    */
   async main() {
     const measurables = await this.api.measurables();
```
OAGr: It's important to only get measurables that have competitive entrees that are newer than aggregates by this specific bot. Can that be added here, or with new stories?

GalochkinKirov: I think would be better in another issue

OAGr: ok, plz review this new issue.
https://github.com/OAGr/foretold/issues/187

bots/bots/aggregation-bot.js
GalochkinKirov: Fixed.

PR #190, opened 2019-05-05
Added py_test - an entry point for running the cdf code from a jupyter notebook

drazenz:
And fixed two bugs in sampling and resampling a cdf


OAGr:
Looks good, thanks! 
Can you please add a few comments in the py_test file to describe what it is & how to use it?

Issue #176, opened 2019-04-25
CDF Library: Turn into separate node library, with Lerna


OAGr:
Rather than being part of the server, the CDF library should be it's own library, with a separate package.json file. Tests should run. 

The application should be set up with Lerna for this to work smoothly. Eventually, the library will have to also be imported into the client. 

https://github.com/lerna/lerna

PR #198, opened 2019-05-07
Feature/#176

GalochkinKirov:
https://github.com/OAGr/foretold/issues/176


bots/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "bots",
+  "name": "@foretold/bots",
```
OAGr: The bots should import the ``cdf`` package, which is needed. (Right now I believe it's manually copied & pasted, which is exactly what Lerna should help with)

GalochkinKirov: Fixed.

GalochkinKirov:
Fixed.

PR #244, opened 2019-05-22
Footer component

goodcodedev:
Added a commit with improvements and right user :)

### DIFF
--- a/packages/components/example/Example/ExampleFullPage.re
+++ b/packages/components/example/Example/ExampleFullPage.re
@@ -69,4 +69,5 @@ let make =
         <Div flex=2> Example__MeasurableTopCard.make </Div>
       </Div>
     </Div>
+    <Footer/>
   </Div>;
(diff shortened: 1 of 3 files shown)


OAGr:
Nice work with this! 
A few comments:

- The ``margin2`` and ``padding2`` functions should really have a (~v) and (~h) tag, like
```
    [
      margin2(~v=`em(2.), ~h=`zero),
      padding2(~v=`zero, ~h=`em(2.))
    ]
```

- The links on the bottom of it should be actual links, so we know that styling on links works.
- The components itself shouldn't have any information about Foretold in it. There should be some way for it to be configured in the examples, similar to how the other examples work.
- The media query string, ``(min-width: 720px)``, should probably be in a configurating file.

OAGr:
In my editor all the margin2 elements without the ``~v/~h`` tags get a warning. You may want to install one of the ReasonML plugins to help here.
image

PR #245, opened 2019-05-24
Adjustments to footer.

goodcodedev:
Content is now passed to the component.
Attempted to do this through a tuple with element and array of items, thought it was clean syntax wise and little runtime cost.
Also added a color, linkAccent, which I think could be valuable for some higher contrast links.


OAGr:
Code looks good!

PR #251, opened 2019-05-29
Feature/component lib

goodcodedev:
Did a quick wrap on this. Think it's something usable, can put in components to understand and work on single component.

Added script entry (yarn) `complib` besides `dev` to not break current behaviour.
With `yarn complib`, example/index.html will be accessible from localhost:1234/index.html, and component viewer will be on localhost:1234/complib/complib.html.

The components definitions is in Complib.re.


OAGr:
Sweet, happy to see this, nice work!
I haven't gone through the code itself much, but did play with it a bit in the browser.

OAGr:
A few small changes:
1. Could you rename complib to something like, "showcase" or "styleguide" maybe?
2. Can you add information on how to use the component viewer in the component README? I'll definitely forget that url. Better yet may be to have a page at index.html or whatever with links to the other two pages.

goodcodedev:
Renamed to showcase and added to readme.

I had problems getting a page to link to the two working with parcel. If you are content without, you could save some brain energy not reading the below :) 

Details:
It was actually quite tricky to make an index link to the two. First, assuming localhost/example/index.html, gave problems with parcel cache if the page url varied (needed rm -rf .cache). This solution could possibly also give problems with assets/links.

Trying to get example/index.html to be available on /index.html sort of worked, but relied on a bit random behavior of parcel, but lastly, even if parcel would serve something like intro.html on `/` when specified first, this would fail because of /index.html link which parcel would try to resolve relatively on the file system.. :) I tried setting the href via javascript as maybe the last resort at this point, and somehow now `intro.html` is not served at `/`.

PR #252, opened 2019-05-30
Feature/#241

goodcodedev:
Bindings for rc-dropdown and rc-menu. Some rough edges with styling, don' think there are direct problems otherwise for the simple usage examples in showcase.

### DIFF
--- a/packages/components/package-lock.json
+++ b/packages/components/package-lock.json
@@ -1,0 +1,11378 @@
+{
+    "name": "@foretold/theme",
+    "version": "0.0.1",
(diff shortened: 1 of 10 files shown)


goodcodedev:
In the menu bindings, I added child elements as modules in the main file (FC__Menu). They have slightly generic names, Item, Divider. In examples I have used Menu.() for local open.

OAGr:
Nice work with this. It seems pretty solid from what I can tell so far. Not sure about how difficult it will be to style correctly, but that can be addressed later.

PR #259, opened 2019-06-04
Feature/#238

goodcodedev:
Certainly not perfectly styled, but probably time to check in.
In this PR is:
* Revised "utility classes" for PageCard.Section
 I tried to split style directions into
  * border: `` `top`` | `` `bottom`` | `` `none`` (default)
  * background: `` `white`` (default) | `` `grey`
  * padding: `` `none`` | `` `top`` | `` `bottom`` | `` `all`` (default)
  * flex: true | false
 I used polymorphic variants as they are syntactically quick to write, avoid namespace issues (both border and padding use `` `none``), and are still well typed.
 At first I tried with a list of styles that were straightforwardly mapped to classes, but had some issues like default padding. This later approach seemed safer and easier to me.
 * DropdownMenu and DropdownSelect
 Went a bit further with menu and trigger to support things like select and styling of the menu and trigger. I've built on the css provided by rc-menu/rc-dropdown as well as ant's css.
 The system of rc-menu/rc-dropdown can pretty much support one style which still requires much custom styling. By copying the remaining styles and make part of the styling configurable, we can easily have variations, so currently the menu have the required styling.
 You can see an example of style customization in FC__DropdownMenu.re (the component is for the combination of trigger and menu with styling).
 DropdownSelect builds on DropdownMenu (is used to choose between "CDF", "PDF" on MeasurableForm) and adds state to show selected option, and triggers `onSelect` callback.
* TabList
 A utility class to ease working with list of tabs where one needs to be selected and you want callback with selected tab. It's a bit constraining to only support strings as labels.
 * Alerts
 Made simple alerts ("primary", info, success, warning, error). The colors were copied from bootstrap as I found more equal to the screenshot (vs ants), but they are not equal, I haven't finely adjusted them.
 The screenshot had an exclamation mark icon. Didn't do more on this, was wondering if something like font-awesome might be something to use.


packages/components/src/Base/FC__E.re
```diff
@@ -105,6 +118,31 @@ module L = {
   let update = Rationale.RList.update;
   let iter = List.iter;
   let findIndex = Rationale.RList.findIndex;
+  module React = {
+    let map = (map: 'a => React.element, list: list('a)): React.element => {
+      let a = [||];
+      let rec loop = list =>
+        switch (list) {
+        | [first, ...rest] =>
+          let _ = Js.Array.push(map(first), a);
```
OAGr: why do you need to use Js.Array.push? Can you use some other append method? Can you make sure it's part of E.A instead of calling it directly?

packages/components/src/Base/FC__E.re
```diff
@@ -105,6 +118,31 @@ module L = {
   let update = Rationale.RList.update;
   let iter = List.iter;
   let findIndex = Rationale.RList.findIndex;
+  module React = {
+    let map = (map: 'a => React.element, list: list('a)): React.element => {
```
OAGr: This function seems suspicious to me. Can you not do it with something like ``List.map`` or one of the Rationale List methods?

packages/components/src/Base/FC__E.re
```diff
@@ -105,6 +118,31 @@ module L = {
...
+      let rec loop = list =>
+        switch (list) {
+        | [first, ...rest] =>
+          let _ = Js.Array.push(map(first), a);
+          loop(rest);
+        | [] => React.array(a)
+        };
+      loop(list);
+    };
+    let mapi =
```
OAGr: This one too seems suspicious.

packages/components/src/Base/FC__Icon.re
```diff
@@ -1,6 +1,5 @@
-
 module Questionmark = {
-  let component = ReasonReact.statelessComponent(__MODULE__);
+  let component = ReasonReact.statelessComponent(__MODULE__ ++ "QuestionMark");
```
OAGr: Please add a space, like, " QuestionMark".

packages/components/src/Base/FC__Menu.re
```diff
@@ -16,14 +14,210 @@ external rcDividerClass: ReasonReact.reactClass = "Divider";
...
 [%bs.raw {|require("rc-menu/assets/index.css")|}];
 
+module Styles = {
+  open Css;
+  let menuBorderRadius = `px(3);
+  // Menu general, applied to menu <ul> elements
+  global(
+    ".ft-menu-general, .ft-submenu-general>ul",
+    [
+      fontFamily("Lato"),
```
OAGr: Please move "Lato" to the colors file or another config, and then reference it here and in other places. We shouldn't have it in so many places. We should only have this string in one place.

packages/components/src/Base/FC__Menu.re
```diff
@@ -16,14 +14,210 @@ external rcDividerClass: ReasonReact.reactClass = "Divider";
 
 [%bs.raw {|require("rc-menu/assets/index.css")|}];
 
+module Styles = {
+  open Css;
+  let menuBorderRadius = `px(3);
+  // Menu general, applied to menu <ul> elements
+  global(
```
OAGr: Kudos for doing this. It still seems messy (my guess is that it's a difficult area), but let's keep for now and consider improving later.

OAGr:
This seems like a lot of stuff, nice work. I left a few comments. I could imagine being more perfectionist on the dropdown stuff, but we can improve that more later.

packages/components/src/Base/FC__Menu.re
goodcodedev: Thanks, it's a bit messy and intrinsically messy. I've gone over it some times until being ensured everything was needed and at a somewhat right place. Still not sure of all styling needs, and was assuming some things could be switched between "global/general" styles and specific styles as style-needs become clearer.

OAGr: sounds good

Issue #262, opened 2019-06-09
Agent Prediction Page


OAGr:
Currently the agent predictions page is pretty bad. It splits things up into measurables, which I don't want. 

Instead, have a list of all predictions by the agent, in order. It should look like the previous view, except there should be one new column for the "Measurable Name", which would be a link for the relevant measurable.

We'll add pagination and other niceties in future stories.

PR #280, opened 2019-06-14
Imprrovements/262

GalochkinKirov:
https://github.com/OAGr/foretold/issues/262


packages/client/src/graphql/Foretold__GraphQL__GetAgent.re
```diff
@@ -102,7 +100,7 @@ module Query = [%graphql
 
 module QueryComponent = ReasonApollo.CreateQuery(Query);
 
-let toMeasurables = (measurements: array(node)) => {
+let toMesuarements = (measurements: array(node)) => {
```
OAGr: ``toMeasurements`` (spelled wrong)

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
```diff
@@ -0,0 +1,183 @@
...
+               }
+            }
+        }
+    }
+  |}
+];
+
+module QueryComponent = ReasonApollo.CreateQuery(Query);
+
+let toMesuarements = (measurements: array(node)) => {
```
OAGr: fix spelling here too

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
```diff
@@ -0,0 +1,183 @@
+type measurable = {
+  id: string,
+  name: string,
+  state: Context.Primary.MeasurableState.t,
+  stateUpdatedAt: option(MomentRe.Moment.t),
+  expectedResolutionDate: option(MomentRe.Moment.t),
+};
+
+// @todo:
```
OAGr: what's the plan for this? To do in future PRs, or should we make an issue?

OAGr: Also, note that we have a ``connections`` type in ``Context__Primary``. Could you use that maybe?

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
```diff
@@ -0,0 +1,183 @@
...
+  |> Array.map(e => e##node)
+  |> E.A.O.concatSome;
+};
+
+module Query = [%graphql
+  {|
+    query getAgentMeasurements ($id: String!, $first: Int, $last: Int, $after: String, $before: String) {
+        agent:
+        agent(id: $id) @bsRecord{
+            measurements: Measurements (first: $first, last: $last, after: $after, before: $before) {
```
OAGr: Instead of doing this, can you use the ``measurements`` query, where you pass in the ``agentId`` (instead of using the agent > measurements query)? That would allow us to filter on more things in the future, unless we want to add the same filtering to this subquery.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
```diff
@@ -0,0 +1,183 @@
...
+type inputType('a) =
+  (
+    ~id: string,
+    ~first: int=?,
+    ~last: int=?,
+    ~after: string=?,
+    ~before: string=?,
+    unit
+  ) =>
+  'a;
```
OAGr: In future stories, we should probably focus more on drying some of this stuff up.

packages/client/src/pages/Agent/AgentShow.re
```diff
@@ -1,56 +1,80 @@
...
-      {
-        r.competitorType
-        |> Context.Primary.CompetitorType.toString
-        |> ste
-        |> E.React.inH3
-      }
-    </>
-  | {user: Some(r)} => r.name |> ste |> E.React.inH1
-  | _ => notFound
+module ReducerConfig = {
```
OAGr: I think we should actually rename this page from AgentShow -> AgentMeasurements, which is more accurate.

packages/client/src/pages/Agent/AgentShow.re
```diff
@@ -1,56 +1,80 @@
...
+                e.id,
+              );
+            };
+
+            let measurementsList = connection.edges |> Array.to_list;
+
+            C.Measurements.Table.makeAgentPredictionsTable(
+              ~measurementsList,
+              ~onSelect=onSelectClb,
+              (),
```
OAGr: I've noticed when scrolling through these that the order is wrong. When I click on one measurement, and then click the "next" icon on the top right, it doesn't bring me to the measurement/measurable that was next in the table. If it's an easy fix, I recommend making it.

packages/client/src/pages/Agent/AgentShow.re
```diff
@@ -1,56 +1,80 @@
...
-  | _ => notFound
+module ReducerConfig = {
+  type itemType = Context.Primary.Measurement.t;
+  type callFnParams = string;
+
+  let getId = (e: Context.Primary.Measurement.t) => e.id;
+  let callFn = (e: callFnParams) =>
+    Foretold__GraphQL.Queries.AgentMeasurements.componentWithMeasurementConnection(
+      ~id=e,
+    );
```
OAGr: Kudos for adding the reducer to this! I wasn't expecting that this issue but it is a nice touch.

OAGr:
That was pretty decently sized, nice work. I left a few comments of things to fix.

packages/client/src/graphql/Foretold__GraphQL__GetAgent.re
GalochkinKirov: Fixed.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: Fixed.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: I have added description to each todos.  At this moment I do not properly understand how to use "Context.Primary.Connection" module in a right way, this is why I added todos. It need more time to think it over.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: Good idea, will do.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: Ok, got it.

packages/client/src/pages/Agent/AgentShow.re
GalochkinKirov: Yes, it has a sense.

GalochkinKirov: Fixed.

packages/client/src/pages/Agent/AgentShow.re
GalochkinKirov: Sure, thanks.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: Fixed.

packages/client/src/pages/Agent/AgentShow.re
GalochkinKirov: Fixed.

packages/client/src/graphql/Foretold__GraphQL__GetAgentMeasurements.re
GalochkinKirov: Fixed.

GalochkinKirov:
Fixed.

PR #281, opened 2019-06-14
Add multimodal support w/ unit test & shorthand for common distributions

lagerros:
**Background**

Allows the user to write things like: `=multimodal( normal(10,1), AK, lognormal(5,2), [0.6, 0.2, 0.2] )` to get a distribution with multiple modes/peaks, as a weighted combination of other distributions or input cells (`AK` is an arbitrary input cell, Guesstimate style). 

Also allows shorthands, like: `=mm( n(10,1), AK, logn(5,2), [0.6] )`.

Supports a flexible number of input distributions, modulo performance. 

More details documented in the code. 

___

**Testing**

I've tested this using the Guesstimator engine in Foretold (the test file is located in `./samplers/simulator-worker/tests/multimodal_tests.js`). 

I haven't tested with the entire Foretold application, nor with the Guesstimator engine as it currently looks in the Guesstimate app.


OAGr:
nice work getting this up so quickly!

What would you think about wrapping both sets of inputs with arrays? It's pretty unusual in functions for it to work like that.

For instance,
``multimodal( [normal(10,1), AK, lognormal(5,2)], [0.6, 0.2, 0.2])``
instead of 
``multimodal(normal(10,1), AK, lognormal(5,2), [0.6, 0.2, 0.2])``

OAGr:
There's a lot of new input names included in this, as the shorthand. Those are tricky, because they may need to be maintained (though with Foretold less than Guesstimate, because they quickly get converted to distributions, and the text isn't really used after). 

Do you think that "bi" is used much? I think I'd prefer to keep the API simpler. Would probably prefer to encourage people to use multimodal most of the time, instead of bimodal. Don't mind the "bimodal" command, but "bi" is pretty generic.

lagerros:
Removing bi seems fine, but I'm a bit unclear about what the worry is - -
future name collisions?

As for more wrapping, I considered it but preferred to avoid as it's a
trivial inconvenience for the user. (Especially as weight array is
optional.) Though this matters more for use via the interface where you'd
repeatedly type this than for bots, where you'd just type the function
once.

OAGr:
I'm going to merge so we can start using. We can think more about the name conventions later on. As long as we don't spread these too widely, changing later is cheap.

Issue #285, opened 2019-06-17
Figure out why Auth0 keeps logging people out


OAGr:
I get logged out of Foretold every hour or so, it seems. I'm not sure why.

PR #289, opened 2019-06-19
Fix auth0 expiration time out

GalochkinKirov:
https://github.com/OAGr/foretold/issues/285

### DIFF
--- a/packages/client/src/contexts/Context__Auth.re
+++ b/packages/client/src/contexts/Context__Auth.re
@@ -138,20 +138,18 @@ module ServerJwt = {
 };
 
 module CallbackUrlToAuth0Tokens = {
-  open Belt;
   open Utils;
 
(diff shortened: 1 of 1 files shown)


OAGr:
This fix seems a bit hacky to me. The ``matchExpiresIn`` part of the url already seems to be equivalent to 24 hours from what I can tell. Why do you think it needs to be overridden?

GalochkinKirov:
1. The main reason why we store "expires_at" timestamp is that we want to prevent getting backend JWT errors before it happens. I think this is why there method "isObsolete" is used.
2. Auth0 gives us JWT token only for a day or less. And this is right. But since we have our own JWT token (Foretold) we should not get renewed tokens each day.

OAGr:
Makes sense, thanks.

Issue #278, opened 2019-06-13
"Edit Bot" page


OAGr:
Create a new page to allow users to edit their bots.

- This page should use the same form as the "Add Bot" page. (They can edit the name & description).
- When it saves, it should redirect to the current user's "bots" page.
- There should be an "edit" link on the user's "bots" page, that is only visible when that user is viewing their own page.

PR #303, opened 2019-06-25
Feature/278 2

GalochkinKirov:
https://github.com/OAGr/foretold/issues/278


packages/client/src/components/MutationForm.re
```diff
@@ -10,9 +10,23 @@ module Make = (Config: Config) => {
         (),
       ) =>
     switch (result) {
-    | Loading => "Loading" |> ste
+    | Loading => <SLayout.Spin />
```
OAGr: Thanks for adding this improvement here!

packages/client/src/pages/Bots/BotForm.re
```diff
@@ -0,0 +1,93 @@
...
+};
+
+module Form = ReFormNext.Make(FormConfig);
+
+let withForm =
+    (onSubmit, bot: option(Context.Primary.Bot.t), innerComponentFn) => {
+  let initialState: FormConfig.state =
+    switch (bot) {
+    | Some(bot) => {
+        name: bot.name |> Rationale.Option.default(""),
```
OAGr: Can you use ``E.O.default`` or similar instead? I'd prefer not to reference Rationale directly, because it's confusing with three different libraries (Rationale, Belt, StdLib)

packages/client/src/pages/Bots/BotForm.re
```diff
@@ -0,0 +1,93 @@
...
+        ),
+      |]),
+    innerComponentFn,
+  )
+  |> E.React.el;
+};
+
+let onSuccess = (loggedInUser: Context.Primary.User.t, ()) => {
+  Context.Primary.User.getAgent(loggedInUser, agent =>
+    Context.Routing.Url.push(Agent({agentId: agent.id, subPage: AgentBots}))
```
OAGr: kudos for having it route to the right place.

packages/client/src/pages/Entity/EntityIndex.re
```diff
@@ -6,27 +6,32 @@ module Columns = {
...
+
+  let instanceOf: column =
+    Table.Column.make(
+      ~name="Instance Of" |> ste,
+      ~render=(r: record) => r |> EKen.Thing.getInstanceOfName |> ste,
+      (),
+    );
+
+  let idColumn: column =
+    Table.Column.make(
```
OAGr: I like this refactor, it makes sense as a next step.

Issue #299, opened 2019-06-25
Show valueText in measurements


OAGr:
Under the "prediction value", show the valueText. Only do this if the value is a distribution, not if it is a point.

PR #304, opened 2019-06-26
Feature/299

GalochkinKirov:
https://github.com/OAGr/foretold/issues/299


packages/client/src/components/Foretold__Components__Measurements__Table.re
```diff
@@ -91,6 +97,17 @@ module Helpers = {
     | _ => None
     };
 
+  let getValueText = (measurement: measurement) =>
+    switch (measurement.value) {
+    | Belt.Result.Ok(`FloatCdf(_r)) =>
```
OAGr: wh do you need to add "Belt.Result"; rather than just saying "Ok(..."?

GalochkinKirov: Thanks, fixed.

Issue #305, opened 2019-06-27
New Question should allow for a valueType of percentage 


OAGr:
Below "Question Type", in the "Simple" form, there should be a new radio button. One option would say "distribution", the other would say "binary." Selecting these would set the ``valueType`` as ``FLOAT`` or ``PERCENTAGE``, respectfully. 

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/1f06f175-9f54-4e5d-9000-b4c2396e2343)

PR #313, opened 2019-06-28
Feature/305

GalochkinKirov:
https://github.com/OAGr/foretold/issues/305

Please, consider this PR after all previous.

### DIFF
--- a/packages/client/src/pages/Measurable/MeasurableEdit.re
+++ b/packages/client/src/pages/Measurable/MeasurableEdit.re
@@ -40,10 +40,11 @@ module WithEditMutation = {
             expectedResolutionDate |> Js.Json.string |> Rationale.Option.some,
           "resolutionEndpoint": resolutionEndpoint |> Rationale.Option.some,
           "labelSubject": labelSubject |> Rationale.Option.some,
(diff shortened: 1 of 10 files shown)


GalochkinKirov:
Screenshot_20190628_095232

Page example.

packages/client/src/pages/Measurable/MeasurableEdit.re
```diff
@@ -25,23 +25,26 @@ module WithEditMutation = {
...
     let date = showDescriptionDate == "TRUE" ? labelOnDate : "";
     let m =
       GraphQL.make(
         ~id,
         ~input={
           "name": name,
-          "labelCustom": labelCustom |> E.O.some,
-          "labelProperty": labelProperty |> E.O.some,
-          "labelOnDate": date |> Js.Json.string |> E.O.some,
+          "labelCustom": labelCustom |> Rationale.Option.some,
```
OAGr: Why did you change this? I'm trying to stay away from the use of ``Rationale`` directly, unless we want to dramatically change use of the E library.

GalochkinKirov: Because it is not readable.

Issue #325, opened 2019-07-01
Import nickname -> username, email, and picture from auth0 on signup.


OAGr:
When a user signs up, ping the auth0 API, get their nickname/email/picture, and save them for that user. This will require making a new picture field in the database, which will be a string.

If the username fails the validations, then don't save the username.

PR #364, opened 2019-07-08
Feaure/325

GalochkinKirov:
By mistake have done this issue first
https://github.com/OAGr/foretold/issues/325


packages/server/migrations/201907050948001-alter-users-table.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
+  up: async function (queryInterface, Sequelize) {
+    // Columns
+    await queryInterface.addColumn('Users', 'email', {
+      type: Sequelize.STRING(64),
+      allowNull: true,
+    });
+    await queryInterface.addColumn('Users', 'picture', {
+      type: Sequelize.STRING(128),
```
OAGr: why have these restrictions here? (64 / 128)? I could imagine that we may want to increase these numbers later on, especially for picture, where users may want to enter a long URL.

GalochkinKirov: We should apply restrictions on DB always. Though we could increase this limit.

Issue #352, opened 2019-07-04
Users preferences


RomanGalochkin:
Add users preferences:
- "Do not send me emails", default "false".

PR #370, opened 2019-07-09
Feature/352

GalochkinKirov:
#352


GalochkinKirov:
Screenshot_20190709_115958

packages/client/src/pages/Preferences.re
```diff
@@ -0,0 +1,141 @@
...
+      };
+
+  let set: type value. (state, field(value), value) => state =
+    (state, field, value) =>
+      switch (field) {
+      | Emails => {...state, emails: value}
+      };
+};
+
+module Form = ReFormNext.Make(FormConfig);
```
OAGr: Kudos for using ReFormNext here, instead of ReForm

OAGr:
I'm not sure I would have created a whole new table/page yet for this one feature, but it will be useful later. Also, I don't think "emails" is the best column name, but we can change that later.

Issue #354, opened 2019-07-04
Notification (email) sub-system


RomanGalochkin:
For instance, the data flow:
1. "Measurable state is changed" event.
2. Observer for this event create notification type "email" with filled template and subject address. 
3. Consumer of email notifications sends emails.

- Notifications sub-system:
-- Mail types (js dictionary) --> email templates,
-- Event type --> mail type map dictionary,
-- Mail templates (table, model),
-- Notifications (table, model) --> id, notification_type (email), event_name, notification (json), sent_at (datatime),
- Events sub-system (events dictionary, observers),
- Mail type notifications consumer

PR #377, opened 2019-07-10
Feature/354

GalochkinKirov:
#354


packages/server/migrations/201907100048002-templates.js
```diff
@@ -0,0 +1,41 @@
...
+  up: async function (queryInterface, Sequelize) {
+    try {
+      await queryInterface.createTable('Templates', {
+        id: {
+          allowNull: false,
+          primaryKey: true,
+          type: Sequelize.UUID
+        },
+        name: {
+          type: Sequelize.STRING(127),
```
OAGr: do you have any specific reason to make this 127? Is this much better than a regular ``Sequelize.STRING``?

packages/server/migrations/201907100048002-templates.js
```diff
@@ -0,0 +1,41 @@
...
+          allowNull: false,
+        },
+        envelopeTemplate: {
```
OAGr: I’m not familiar with the phrase “envelope” here. Is this basically an email template, or a general message template?

packages/server/migrations/201907100048002-templates.js
```diff
@@ -0,0 +1,41 @@
+module.exports = {
+  up: async function (queryInterface, Sequelize) {
+    try {
+      await queryInterface.createTable('Templates', {
```
OAGr: Can you be more specific than “Template”? Is this a “NotificationTemplate” or an “EmailTemplate”?

packages/server/src/async/consumers/consumer.js
```diff
@@ -0,0 +1,30 @@
+const assert = require('assert');
+const _ = require('lodash')
+
+const data = require('../../data');
+
+const { TEMPLATE_NAME } = require('../../models/enums/template-name');
+const { NOTIFICATION_TYPE } = require('../../models/enums/notification-type');
+const { EmailEnvelope } = require('../../models/classes/notifications');
+
+class Consumer {
```
OAGr: Consumer seems kind of vague. Maybe “NotificationConsumer” or similar?

packages/server/migrations/201907100048002-templates.js
GalochkinKirov: I know that default length in Sequelize.STRING is 255, it is too long for this reason. We just want to give readable names for templates, no more, this is why.

packages/server/src/data/classes/filter.js
```diff
@@ -0,0 +1,34 @@
+const _ = require('lodash');
+
+class Filter {
```
OAGr: For a lot of these, I’d like comments on the top describing what they are for. It’s not obvious to be from a quick glance what this is filtering.

packages/server/migrations/201907100048002-templates.js
GalochkinKirov: The field "envelopeTemplate" is used to save templates for any systems needs, so, the better to rename it into something else here, for instance "template", but "template" is reserved term for model name, so to make difference just let's name it as "envelopeTemplate".

packages/server/migrations/201907100048002-templates.js
GalochkinKirov: It is not only for "notification" or "emails" templates, it is common way of storing "system templates", do not connect it with "notifications" or "emails".

packages/server/src/async/consumers/consumer.js
GalochkinKirov: Here "consumer" class is parent class of all another "consumers", for instance "EmailsConsumer", "PushConsumer" and another ones.

packages/server/src/data/classes/filter.js
GalochkinKirov: Yes, thanks will add some comments, it has a sense.

GalochkinKirov: Have added comments.

packages/server/src/async/consumers/consumer.js
OAGr: Can you write these comments in the code, instead of in replies? That makes sense to me, but I will forget.

packages/server/src/async/consumers/consumer.js
```diff
@@ -0,0 +1,30 @@
...
+  constructor(options = {}) {
+    assert(_.isObject(options), 'Options is not an object');
```
OAGr: Why this pattern of using assert outside a test? I’m not used to that.

packages/server/src/async/consumers/consumer.js
GalochkinKirov: Sure.

packages/server/src/async/consumers/consumer.js
GalochkinKirov: Assertion is not only about tests. Assertion is useful to prevent development and runtime errors.

packages/server/src/models/series.js
```diff
@@ -3,61 +3,60 @@ const Sequelize = require('sequelize');
...
+      type: Sequelize.VIRTUAL(DataTypes.INTEGER),
+      get: getMeasurableCount,
+    },
+  });
+
+  Series.addHook('afterCreate', async (series) => {
+    await series.createMeasurables();
+  });
+
+  // TODO: These queries are likely very slow,
```
OAGr: Kudos for noticing and adding this.

PR #390, opened 2019-07-18
Refactoring

RomanGalochkin:
Sorry for name of branch "feature/322", should have another name, here I was going to solve 322 issue. But for force schedule I need to switch on another branches.


OAGr:
looks good to me

Issue #388, opened 2019-07-18
User pages should show their brier score, for all resolved percentage questions they forecasted


OAGr:
The user brier score should be calculated, and shown on their user page (under the description, don't worry about making it look good)

First pass, this can be calculated on-demand. Later we can cache many of the pieces.

see the comment to:
https://github.com/OAGr/foretold/issues/328

PR #393, opened 2019-07-18
Feature/388

RomanGalochkin:
#388

PR to "epic/emails" to see changed lines.

### DIFF
--- a/packages/server/src/lib/brier-score.js
+++ b/packages/server/src/lib/brier-score.js
@@ -38,15 +38,6 @@ class BrierScore {
    * @return {number}
    */
   _brierScore(probability, result) {
-    if (probability === BrierScore.PERCENTAGES_50) {
-      return BrierScore.REGARDLESS_SCORE;
-    }
-    if (probability === BrierScore.PERCENTAGES_100) {
-      return result === true
-        ? BrierScore.BEST_SCORE
-        : BrierScore.WORST_SCORE;
-    }
-
     const score = result === true
       ? ((probability / 100) - 1) ** 2
       : (probability / 100) ** 2;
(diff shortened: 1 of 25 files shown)


packages/server/src/lib/brier-score.js
```diff
@@ -0,0 +1,66 @@
...
+   */
+  _brierScore(probability, result) {
+    if (probability === BrierScore.PERCENTAGES_50) {
+      return BrierScore.REGARDLESS_SCORE;
+    }
+    if (probability === BrierScore.PERCENTAGES_100) {
+      return result === true
+        ? BrierScore.BEST_SCORE
+        : BrierScore.WORST_SCORE;
+    }
```
OAGr: Why these cases, for PERCENTAGES_50, PERCENTAGES_100? So we wouldn’t have to do those calculations? This seems pretty unnecessary to me; it could speed things up, but these are very unlikely situations.

packages/server/src/models-abstract/measurement-model.js
```diff
@@ -17,6 +18,112 @@ class MeasurementModel extends ModelPostgres {
...
+             "Measurements"."agentId",
+             array_agg("Measurements"."value" ->> 'data') as "probabilities"
+      FROM "Measurements"
+               LEFT JOIN "Measurables"
+                         ON "Measurements"."measurableId" = "Measurables".id
+      WHERE "Measurables"."state" = 'JUDGED'
+        AND "Measurables"."valueType" = 'PERCENTAGE'
+        AND "Measurements"."competitorType" = 'COMPETITIVE'
+        AND "Measurements"."agentId" = '${agentId}'
+      GROUP BY "Measurements"."measurableId", "Measurements"."agentId"
```
OAGr: Nice work doing all of these queries in raw SQL, that should speed things up.

packages/server/src/lib/brier-score.js
OAGr: I removed this in a commit.

RomanGalochkin: Yes, sure, did not notice it when was writing code, thanks.

Issue #403, opened 2019-07-21
Email: email forecasters when a question is first resolved


OAGr:
When a question is resolved (closed), all people who forecasted that question, besides the person who resolved that question, should be emailed saying as such. This could be pretty simple for a first pass.

"The question [How many bats will be born in 2019] was resolved. Click here to see the result."

(Question name in brackets)

PR #411, opened 2019-07-23
Feature/403

RomanGalochkin:
#403


packages/server/src/models/hooks.js
```diff
@@ -0,0 +1,27 @@
+const events = require('../async/events');
+const emitter = require('../async/emitter');
+
+const { MEASURABLE_STATE } = require('./enums/measurable-state');
+
+function addHooks(db) {
+  db.sequelize.addHook('afterUpdate', (instance) => {
```
OAGr: why are these here instead of in files related to the model, like models/measurable?

OAGr:
There were a whole lot of small changes made for this PR. Can you briefly describe the main reasons for these?

packages/server/src/models/hooks.js
RomanGalochkin: The purpose is to encupsulate this part of application. Here "Models" just drivers to DB nothing more.

RomanGalochkin:
Yes, sure. Previously when we started to implement "models drivers"/"models"/"datas layers" template I have made an error in interfaces definitions, so to repair this situation I need to rename all legacy methods of "data layer" into to something "nameOfMethod2" to use parents good written methods "getOne" and another. This will keep lines of code on the next stages of application live.

Issue #409, opened 2019-07-21
Communities in sidebar, other than "Home", should show in alphabetical order


OAGr:
A on top. This should probably just be done on the client side.

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/f31cfcfc-b4d9-474d-b05e-04d59546cbd1)

PR #420, opened 2019-07-25
Feature/409 2

RomanGalochkin:
#409


packages/client/src/lib/Sorting.re
```diff
@@ -0,0 +1,11 @@
+type sortable = Types.channel;
```
OAGr: I really don't like that this module is called ``sorting`` but it only uses ``Channel`` for some reason. Either figure out some nicer way of doing this (with a Functor, perhaps), or I wouldn't recommend it like this.

This also makes the assumption that the array has items with names that are strings.

Issue #399, opened 2019-07-19
Upgrade dependencies to fix Github Security Alerts


OAGr:
See a private message on what exactly these are. Use your judgement in deciding if some may not be relevant.

RomanGalochkin:
@OAGr Could you give a small explanation how to find "Github Security Alerts"?

OAGr:
I think I sent a message in slack with screenshots. My guess is that you don't have access because of access privileges. 

![image](https://user-images.githubusercontent.com/377065/61939250-c19ab500-af8a-11e9-8c42-2cfe912b4d9f.png)

PR #429, opened 2019-07-29
Bumps versions up

RomanGalochkin:
#399


RomanGalochkin:
btw I am not sure that it will help

Issue #322, opened 2019-07-01
Redirect logged-in user to profile page if their username is blank


OAGr:
This will force them to add a username first thing.

PR #432, opened 2019-07-30
Improvements/322

RomanGalochkin:
#322 

Please, consider previous are not merged PRs.


packages/client/src/Redirect.re
```diff
@@ -0,0 +1,37 @@
+let component = ReasonReact.statelessComponent("Redirect");
+
+// @todo: fix this duplicated ligic(#rederectings)
```
OAGr: Do you have an idea when you are going to do this? Can you make a story for it?

packages/client/src/graphql/Authentication.re
```diff
@@ -44,20 +40,22 @@ let component =
...
+  <QueryComponent variables=query##variables>
+    ...{({result}) =>
+      result
+      |> HttpResponse.fromApollo
+      |> HttpResponse.fmap(e => e##authentication##jwt)
+      |> (
+        e =>
+          switch (e) {
+          | Success(c) =>
+            // @todo: rapir this hard logic
```
OAGr: This too. Maybe some refactoring story or similar?

packages/client/src/App.re
```diff
@@ -1,50 +1,46 @@
 type state = {
   route: Routing.Route.t,
   authToken: option(string),
+  routingCount: int,
```
OAGr: Why is this needed? Can you add a comment to the code or similar? I imagine it could also have a more specific name, I really don't understand what this means.

packages/client/src/Redirect.re
RomanGalochkin: Here I need to rethink the application structure more before doing it. For now I do not properly imagine how to do it. 

But I have added new issue with steps: 
https://github.com/OAGr/foretold/issues/434

packages/client/src/graphql/Authentication.re
RomanGalochkin: Yes, have added new issue: https://github.com/OAGr/foretold/issues/434

packages/client/src/App.re
RomanGalochkin: Thanks for your question, have changed this a bit, moved this var into "Redirect" component and renamed it into "countOfRedicrectings". This var need to prevent "user cannot go from profile page".

OAGr: Maybe change the name to "redirectionCount" or at least "countOfRedirections", to fix spelling?

RomanGalochkin: Thanks, fixed.

Issue #461, opened 2019-08-06
[Bot]: Auto-resolving is not working


RomanGalochkin:

OAGr:
Can you describe what this means? I have no idea, really.

RomanGalochkin:
I have found that there is no bots or something else in the application to resolve Measurables which need to be "resolved after expected resolution date"

OAGr:
Ok, makes sense

OAGr:
Maybe we could change the name to something like: "Bug: Auto-resolving is not working"?

Issue #458, opened 2019-08-05
Add new FeedItem Types


OAGr:
**A new measurement/prediction is made**
'{{ agent.name }} made a prediction'

**A new measurement/comment is made**
'{{ agent.name }} made a new comment'

**A new measurement/resolution(objective answer) is made**
'{{ agent.name }} resolved the question with an answer'

**A new measurement / RESULT_NOT_AVAILABLE is added**
'{{ agent.name }} has marked this question as unresolved'

**A question reaches it's resolution date**
'Question has reached it's expected resolution date'

**A new question is made**
'{{ agent.name }} made a new question'

PR #462, opened 2019-08-06
Feature/458

RomanGalochkin:
#458 

All except this is ready:

> A question reaches it's resolution date
> 'Question has reached it's expected resolution date'

Blocked by #461


packages/server/src/async/producers/notifications/measurable-state-changed.js
```diff
@@ -6,16 +6,38 @@ const { MeasurableState } = require('./measurable-state');
...
+  constructor(measurable) {
+    super(measurable);
     this.templateName = Producer.TEMPLATE_NAME.MEASURABLE_STATE_IS_CHANGED;
   }
 
+  /**
+   * @protected
+   * @return {Promise<boolean>}
+   */
+  async _isActual() {
```
OAGr: Can you describe what this means, perhaps with a better name? ``isActual`` doesn't mean much to me.

RomanGalochkin: When triggers like this "MeasurableStateChanged" are fired first of all they should detect "does an event correspond to them?".

Issue #255, opened 2019-06-01
Move "Guesstimator" library from /packages/client into it's own package


OAGr:
This will be used in the "Components" library shortly after.

It's package should be named "guesstimator"

Also, spend some time cleaning up the interface. Spend up to 2 days on this.

OAGr:
It should be possible to later import the library into Guesstimate, where it originally came from.

https://github.com/getguesstimate/guesstimate-app

RomanGalochkin:
Move the package - 1 SP I think, interfaces - another deal.

PR #470, opened 2019-08-07
Improvements/255

RomanGalochkin:
#255


packages/guesstimator/.flowconfig
```diff
@@ -0,0 +1,17 @@
+[ignore]
```
OAGr: Does it need this? Does it use flow?

RomanGalochkin: Yes, as I see it uses flow.

PR #524, opened 2019-08-12
Improvements/287

RomanGalochkin:
The main part of these changes is:
```
    if (!values || values.length === 0) {
      this.props.onUpdate([[], [], false]);
    } else if (values.length === 1) {
      this.props.onUpdate([[1], values, false]);
    } else {
      const min = this.props.min || parsedInput.params[0];
      const max = this.props.max || parsedInput.params[1];
      this.props.onUpdate(toCdf(values, min, max));
    }
```

So here we see that min & max values which we pass to CDF (pdfast lib) are being set either Measurable.min/max or cdf input field.

### DIFF
--- a/packages/cdf/cdf.js
+++ b/packages/cdf/cdf.js
@@ -1,5 +1,4 @@
 const { interpolate, range, min, max } = require('./functions');
-const { Pdf } = require('./pdf');
 
 class Cdf {
   /**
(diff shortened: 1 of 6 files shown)


OAGr:
Seems good. This part of the code is decently messy, especially for reasons that are my fault (sorry!)

Issue #490, opened 2019-08-08
[Server] Load the Ken Data.json file from the server


OAGr:
There should be a table called ``GlobalSettings`` or similar. Now, it has one column, called ``Entities``. This returns a json object that is the information from Data.json.

This should be available via graphql.

PR #531, opened 2019-08-13
Feature/490

RomanGalochkin:
#490

### DIFF
--- a/packages/server/migrations/201908131532000-table-global-settings.js
+++ b/packages/server/migrations/201908131532000-table-global-settings.js
@@ -9,6 +9,11 @@ module.exports = {
           primaryKey: true,
           type: Sequelize.UUID
         },
+        name: {
+          type: Sequelize.STRING,
+          allowNull: false,
+          unique: true,
+        },
         entity: {
           type: Sequelize.JSON,
           allowNull: true,
--- a/packages/server/migrations/201908131532000-table-global-settings.js
+++ b/packages/server/migrations/201908131532000-table-global-settings.js
@@ -1,0 +1,49 @@
+module.exports = {
+  up: async function (queryInterface, Sequelize) {
+    try {
+      await queryInterface.sequelize.query(`BEGIN`);
+
+      await queryInterface.createTable('GlobalSettings', {
+        id: {
+          allowNull: false,
+          primaryKey: true,
+          type: Sequelize.UUID
+        },
+        entity: {
+          type: Sequelize.JSON,
+          allowNull: true,
+        },
+        createdAt: {
+          type: Sequelize.DATE,
+          defaultValue: Sequelize.NOW,
+        },
+        updatedAt: {
+          type: Sequelize.DATE,
+          defaultValue: Sequelize.NOW,
+        },
+        removedAt: {
+          type: Sequelize.DATE,
+          allowNull: true,
+        },
+      });
+
+      await queryInterface.sequelize.query(`COMMIT`);
+    } catch (e) {
+      console.error('Migration Up Error', e);
+      await queryInterface.sequelize.query(`ROLLBACK`);
+      throw e;
+    }
+  },
+
+  down: async function (queryInterface) {
+    try {
+      await queryInterface.sequelize.query(`BEGIN`);
+      await queryInterface.dropTable('GlobalSettings');
+      await queryInterface.sequelize.query(`COMMIT`);
+    } catch (e) {
+      console.error('Migration Down Error', e);
+      await queryInterface.sequelize.query(`ROLLBACK`);
+      throw e;
+    }
+  }
+};
--- a/packages/server/package-lock.json
+++ b/packages/server/package-lock.json
@@ -4,6 +4,19 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@apollographql/apollo-tools": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@apollographql/apollo-tools/-/apollo-tools-0.4.0.tgz",
+      "integrity": "sha512-7wEO+S+zgz/wVe3ilFQqICufRBYYDSNUkd1V03JWvXuSydbYq2SM5EgvWmFF+04iadt+aQ0XCCsRzCzRPQODfQ==",
+      "requires": {
+        "apollo-env": "0.5.1"
+      }
+    },
+    "@apollographql/graphql-playground-html": {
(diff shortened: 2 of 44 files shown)


packages/server/migrations/201908131532000-table-global-settings.js
```diff
@@ -0,0 +1,54 @@
...
+          defaultValue: Sequelize.NOW,
+        },
+        updatedAt: {
+          type: Sequelize.DATE,
+          defaultValue: Sequelize.NOW,
+        },
+        removedAt: {
+          type: Sequelize.DATE,
+          allowNull: true,
+        },
```
OAGr: why do we need a removedAt?

packages/server/migrations/201908131532000-table-global-settings.js
```diff
@@ -0,0 +1,54 @@
...
+        id: {
+          allowNull: false,
+          primaryKey: true,
+          type: Sequelize.UUID
+        },
+        name: {
+          type: Sequelize.STRING,
+          allowNull: false,
+          unique: true,
+        },
```
OAGr: What is the name? I'm not sure this is needed.

packages/server/migrations/201908131532000-table-global-settings.js
```diff
@@ -0,0 +1,54 @@
...
+        entity: {
+          type: Sequelize.JSON,
+          allowNull: true,
+        },
```
OAGr: If this represents the entirety of data.json, then maybe call this field ``entityGraph`` or similar instead, to make it obvious it's a bunch of entities?

OAGr:
The core functionality looks fine. I don't quite understand why this required so many changes though; were you doing some semi-related refactoring as part of this? Can you explain the changes briefly?

packages/server/migrations/201908131532000-table-global-settings.js
RomanGalochkin: One of the good practice is not to remove objects but just mark them as "removed", so for the future system growing or prevent future changes in this table I would like to add this column. But it is not really necessary and I may remove "removedAt" column. Should I remove column since is not clear why the column had been added?

OAGr: Yea. From what I understand, we'll only really have one row, and won't ever delete it, so I don't think this would be necessary.

RomanGalochkin: May be there will be a need to store something else instead of not just one file? So in this case this table "GlobalSettings" are standing like a common place of storing data.

OAGr: I'm thinking that we may want to also store settings, like, "defaultAggregatorBotId". However, even here there would just be one row.

If we wanted multiple options of Ken items, then we would make a separate ``Entity`` table or similar.

packages/server/migrations/201908131532000-table-global-settings.js
RomanGalochkin: I do think that we need to create this table for a common purpose. Storing one row and one file in the table is bad practice and something strange. One step of extending functionality of the table is to add "name" column to make it possible to identify rows .

packages/server/migrations/201908131532000-table-global-settings.js
RomanGalochkin: Ok, so I am going to remove "removedAt" column according to these comments.

packages/server/migrations/201908131532000-table-global-settings.js
OAGr: Could you give me an example of what a few rows would be? 

Having a one-row table for common app settings, that are going to change (unlike Env vars), seems pretty reasonable to me.

RomanGalochkin: Oh, do you mean storing ALL settings in one row? In one field?

My view is to store settings like this:
"Data.json" --> JSON of the Data.json
"notificationEnabled" --> true
"smtpServer" ---> host (string)
and others, and only after this structure we are able to add for instance "isPublic" column or similar.

So if we add only one row, then to prevent access issues we will need to add some flags right into JSON. For instance, "{dataJson: {}, dataJsonPublic: true, anotherSettings: 1, anotherSetting2: 2} "

RomanGalochkin: And this is only first case where we can fall down. There are 5 rules of DB normalization, and one of them is to put different things in different rows.

OAGr: I mean storing them all in one row, but in multiple fields. So in that example, we'd have a column for each item. 

Can we prevent access issues on a graphQL side? As in, it's easy to get the info from the db, but it won't actually send it to a client?

I'm imagining, similarly, that in the graphQL side, we don't need to know these are rows. We could just have a query for something like,

```graphQL
globalSettings{
  entityGraph
  notificationsEnabled
  smtpServer
}
```

RomanGalochkin: So there are two approaches. You propose to store settings in one row with multiple columns and I propose to store named setting in multiple rows with. Got your point, thanks. Yes, now I  understood that in my approach to get all settings of the application we would send a lot of queries. 


Thanks, going to change it a bit.

packages/server/migrations/201908131532000-table-global-settings.js
RomanGalochkin: Sure, thanks.

packages/server/migrations/201908131532000-table-global-settings.js
OAGr: Sounds good. I'm not really sure, but for one, we could always change later.

RomanGalochkin:
So the latest interfaces of different layers are not in consistency state with legacy ones. So to make small steps to correct this I fixed some of issues. I think that just changing it in one moment is hard thing, so it is easier to make it in small portions. 

Should I repair this PR and create issue with description?

OAGr:
That sounds good to me. I'm fine with making changes in small portions, just am trying to understand what is going on.

packages/server/migrations/201908131532000-table-global-settings.js
RomanGalochkin: Never mind, my misunderstanding.

RomanGalochkin:
Please aim me to the string where you did get what's going on. May be have a sense to add comments or similar.

Issue #492, opened 2019-08-08
[Server] Push from Github repo to server update


OAGr:
There should be a webhook set up; whenever a github repo changes, the server should update the database with the most recent Ken Update.

alex-coding-c:
Hi @OAGr 
Seems you want to integrate CI/CD for the project.
Feel free to check my github profile and I'd like to talk something with you in detail.

Best
Kinji

PR #539, opened 2019-08-15
Feature/492

RomanGalochkin:
#492 

The PR is made into 491 to observe difference. 

For the prod version we should set 
GITHUB_PERSONAL_ACCESS_TOKEN (https://github.com/settings/tokens)
GITHUB_WEBHOOK_SECRET (some random generated string)
GITHUB_REPO_OWNER=foretold-app
GITHUB_REPO_NAME=ken-data
SERVER_URL=https://prediction-backend-staging.herokuapp.com
and for prod
SERVER_URL=https://foretold.io

Entity Graph is being updated after a PR into repository will have been merged.


RomanGalochkin:
For the staging all environments were set except GITHUB_PERSONAL_ACCESS_TOKEN. This var was set to "_", here we need to set your own personal access token @OAGr

RomanGalochkin:
The same was done for the prod version.

When you will create "Personal Access Token" please do not forget to choose "admin:repo_hooks" scope.

docker-compose.override.example.yml
```diff
@@ -4,17 +4,42 @@ services:
...
+      # - JWT_ISSUER=Foretold # 3.6
+      - JWT_SECRET=oDtWHuh0r44y6YSkU0886fVe5br7Thr6 # 3.7
+      # 4. SMTP
+      - SMTP_HOST=email-smtp.us-east-1.amazonaws.com # 4.1
+      # - SMTP_PORT=465 # 4.2
+      - SMTP_USER=someuser # 4.3
+      - SMTP_PASS=somepassword # 4.4
+      - SMTP_FROM=UserName <userbox@host.com> # 4.5
+      # 5. GitHub. You must be owner of the repo.
+      - GITHUB_PERSONAL_ACCESS_TOKEN=K3Gyw6JjxtQyu82iAbzNQ9GDozBESNNVKjtOboW6 # 5.1
```
OAGr: Is this a private token? Is it ok for it to be in a public github repo?

OAGr: I'd suggest we don't have things that even look secretive here. Even if they're actually not, it could raise questions to people reading this. We could deal with these in other ways.

packages/server/src/github/api.js
```diff
@@ -0,0 +1,231 @@
...
+class API {
+  /**
+   * @public
+   */
+  constructor() {
+    this.repoOwner = config.GITHUB_REPO_OWNER;
+    this.repoName = config.GITHUB_REPO_NAME;
+    this.token = config.GITHUB_PERSONAL_ACCESS_TOKEN;
+    this.webhookSecret = config.GITHUB_WEBHOOK_SECRET;
+    this.apiURL = `https://api.github.com`;
```
OAGr: I feel like this should go into a constants file somewhere, at the very least be a constant on the top of the file. I don't like magic numbers.

packages/server/src/github/api.js
```diff
@@ -0,0 +1,231 @@
...
+      console.warn(
+        `GitHub personal access token is not set, ` +
+        `see https://github.com/settings/tokens.`
+      );
+      this.isReady = false;
+    }
+    if (!this.webhookSecret) {
+      console.warn(`GitHub webhook secret is not set`);
+      this.isReady = false;
+    }
```
OAGr: Again, I feel like the top 20 lines could be dried up with a pattern. Like,
``validate(!this.token, "Github personal access token is not set")``

OAGr: On that note, the text messages being here is a bit sloppy. We could have a file or at least a {} lib for text.

``ERROR_TEXT = {
NO_OWNER = `GitHub repo owner is not set.`
...
}

packages/server/src/github/api.js
```diff
@@ -0,0 +1,231 @@
...
+   * @protected
+   * @return {{Authorization: string, Accept: string, "User-Agent": string}}
+   */
+  _getHeaders() {
+    assert(_.isString(this.token),
+      'GitHub personal access token should be a string');
+    return {
+      'Authorization': `bearer ${this.token}`,
+      'User-Agent': 'Foretold App',
+      'Accept': 'application/vnd.github.v3+json',
```
OAGr: I'd also want these to be pulled out

packages/server/src/github/trigger.js
```diff
@@ -0,0 +1,62 @@
...
+   * @param {object} webhook
+   */
+  constructor(webhook, xHubSignature) {
+    this.webhook = webhook;
+    this.xHubSignature = xHubSignature;
+    this.api = new API();
+    this.globalSetting = new GlobalSettingsData();
+
+    assert(_.isObject(webhook), 'WebHook should be an object');
+    assert(_.has(webhook, 'pull_request.merged'), 'WebHook is not in format #1');
```
OAGr: I don't like the ``format #1`` phrase. Maybe something like,
``Webhook does not have required pullrequest.merged attribute. This may be because it's in the wrong format.``

packages/server/src/github/trigger.js
```diff
@@ -0,0 +1,62 @@
...
+      console.warn('Pull Request number is required.');
+      return false;
+    }
+
+    const dataJson = await this.api.getDataJson(pullRequestNumber);
+    if (dataJson) {
+      console.log('Data.json', dataJson);
+      await this.globalSetting.updateEntityGraph(dataJson);
+    } else {
+      console.warn(`Data.json file is not found in the PR ${pullRequestNumber}`);
```
OAGr: kudos for specifying a bunch of possible errors

OAGr:
This seemed decently messy, kudos for getting all of it work successfully. I'm looking forward to trying & using it.

I left a few comments, please review.

packages/server/src/github/api.js
RomanGalochkin: Do you mean the last string with GitHub URL?

packages/server/src/github/api.js
RomanGalochkin: What exactly here?

packages/server/src/github/api.js
OAGr: yep

packages/server/src/github/api.js
OAGr: Sorry; the most important bits are the "Foretold App" and 'application/vnd.github.v3+json' strings

packages/server/src/github/api.js
```diff
@@ -0,0 +1,231 @@
...
+
+    const options = {
+      uri,
+      method,
+      body,
+      headers: this._getHeaders(),
+      json: true,
+      followAllRedirects: true
+    };
+    // console.log('GitHut query options', options);
```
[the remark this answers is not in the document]
RomanGalochkin: I think in the future I will uncomment this string not a once, but we could remove it of course.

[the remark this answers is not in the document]
OAGr: ok, up to you

packages/server/src/github/trigger.js
RomanGalochkin: Sure, thanks, it was delivered to later thinking.

RomanGalochkin:
Thanks, will fix these things tomorrow, it won't take much time.

OAGr:
sg!

docker-compose.override.example.yml
RomanGalochkin: I had an idea that "example" file should show particular samples of values but not just "<change-me>". When one sees string with an example of value he better understand what this environment means. Sure, there were not real tokens.

OAGr: Ah. In that case could you make them obviously fake?

RomanGalochkin: Great idea, will do

packages/server/src/github/api.js
RomanGalochkin: This messages are aimed only for developers, it is not a public part of the application. So this is why here is no need to put it to somewhere to place it in one place. These part of code is stuff. Today this part is tomorrow is not. So I believe that the better approach here is to "inline" code instead of using magic and help functions.

docker-compose.override.example.yml
RomanGalochkin: Done.

packages/server/src/github/api.js
RomanGalochkin: Have added a little refactoring here.

RomanGalochkin: So even If I use a dictionary with these messages for developers I will have to do it over the whole application to make it consistency and do not confuse developers.

packages/server/src/github/trigger.js
RomanGalochkin: Fixed.

packages/server/src/github/api.js
RomanGalochkin: Fixed.

packages/server/src/github/api.js
RomanGalochkin: Fixed.

packages/server/src/github/api.js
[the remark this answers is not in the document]
RomanGalochkin: Yes, it looks messy with commented code, so removed that line.

RomanGalochkin:
I think all what I planed to fix I had fixed. Thanks a lot for your comments it is important to keep communication between developers.

packages/server/src/github/api.js
OAGr: I'd agree you'd want consistency, which is why I recommended eventually just making a new story/chore for doing this in a broader way.

RomanGalochkin: Anyway it is not ordinary way to store dev messages in one place, need to review another applications to make decision.

RomanGalochkin: And another argument that it will raise cohesion between encapsulated modules (https://en.wikipedia.org/wiki/Cohesion_(computer_science)), so let's stay it this way may be for a short time.

Issue #533, opened 2019-08-13
Client should either refresh jwt or force user to re-log-in after jwt expires


OAGr:
I tried logging in locally, and got this error in the server. On the client I didn't see an error; I just saw the logged-out home page without any error.

In order to fix it, I deleted my local application storage for the app.

I imagine this may happen in production, but am not sure.

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/70847c6c-6dd2-47fe-bbdb-abd813b219e3)

This could also be an opportunity to refactor this part of the codebase a bit.

PR #548, opened 2019-08-19
Improvements/533

RomanGalochkin:
#533


RomanGalochkin:
Along with the PR I think we should update Auth0 configuration to support redirect logout link (/link).

OAGr:
Do you mean change settings in auth0?

RomanGalochkin:
Yes, are there something similar?

OAGr:
K, will do so later.

Issue #544, opened 2019-08-16
All public activity shows up in the home channel.


OAGr:
I'm seeing measurables from channels that I am not in, in my home menu.

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/2a96e328-3b53-427d-90cd-445cfe724960)

Here, it shows items in ``#X-risk predictions``, even though I'm not in that channel.

PR #561, opened 2019-08-20
Improvements/544

RomanGalochkin:
#544

### DIFF
--- a/packages/server/src/lib/utils.js
+++ b/packages/server/src/lib/utils.js
@@ -9,7 +9,7 @@ const _ = require('lodash');
  */
 module.exports.diff = function diff(className, options, list) {
   const difference = _.difference(_.keys(options), list);
-  if (difference > 0) {
+  if (difference.length > 0) {
     console.warn(`${className} don't expect "${difference.join(', ')}".`);
   }
   return difference;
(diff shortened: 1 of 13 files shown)


packages/server/src/lib/utils.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+};
+
+/**
+ * @param className
+ * @param list
+ * @param options
+ * @param context
+ */
+module.exports.extend2 = function extend2(className, options, list, context) {
```
OAGr: can there be some better name then ``extend2``, ``diff2`` and ``test2``. The 2 doesn't mean anything to me.

packages/server/src/lib/utils.js
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * @param className
+ * @param options
+ * @param list
+ * @return {string[]}
+ */
+module.exports.diff = function diff(className, options, list) {
+  const difference = _.difference(_.keys(options), list);
+  if (difference.length > 0) {
+    console.warn(`${className} don't expect "${difference.join(', ')}".`);
```
OAGr: It should probably be ``doesn't``, not ``don't``

RomanGalochkin: Yes, sorry

packages/server/src/lib/utils.js
RomanGalochkin: Yes, sure

RomanGalochkin:
Updated.

PR #562, opened 2019-08-20
Fixes condition

RomanGalochkin:


RomanGalochkin:
Sorry, was my mistake. Should pay more attention at these things.

RomanGalochkin:
The staging is tested, think all should be ok on the prod version too. But will test it after.

Issue #549, opened 2019-08-19
Server: GraphQL Query for measurementScoreSet


OAGr:
```graphql
type measurementScoreSet {
    prediction: Measurement! (competitive)
    outcome: Measurement! (objective)
    previousAggregate: Measurement (aggregate)
    primaryPointScore: float!
}

type Measurement {
  ...
  meaurementScoreSet: MeasurementScoreSet
}
```

In the cases that measurements have outcomes, they should have ``measurementScoreSets``. This will eventually be used to return all of the possible scores the measurement created.

The ``primaryPointScore`` should return a random number at this point. Feel free to do this via making a new database table or not at this point.

PR #574, opened 2019-08-23
Feature/549

RomanGalochkin:
#549


OAGr:
Does this code actually return the actual measurementScoreSet? The code seems so simple, like the part of the logic to figure out what the outcome measurement and previousAggregate is missing.

RomanGalochkin:
No, it does not return actual data. It is not obvious from the issue how it should work.

OAGr:
Ok, I left comments in slack.

RomanGalochkin:
Have updated this PR.

Issue #554, opened 2019-08-19
Client: New Table of Measurements and Scores


OAGr:
Each community, except for the Home community, should have a "Leaderboard" tab. This will display a list of:

- All competitive measurements of JUDGED measurables.
- For each, show:
  - The measurable name (a link)
  - The agent name (link)
  - The score (this should be random or similar at this point)
  - The time the measurable was made.

Don't worry about showing the Prediction or Resolution Value yet.

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/538fbfb5-3da0-4fa9-8f81-7bbcebc5ba0a)

PR #578, opened 2019-08-26
Feature/554

RomanGalochkin:
#554


OAGr:
In general looks good, but there seems to be a bug.

When I try this locally, and open a question, it gives me a blank page with an error in the console.
image

```
uncaught exception: Match_failure,-7,Foretold__Components__Measurable__FullPresentation.re,79,20
```

RomanGalochkin:
Yes, have found the same bug. Will update in five minutes.

RomanGalochkin:
Fixed. Thanks.

Issue #579, opened 2019-08-26
Hide leaderboard tab unless EnableExperimentalFeatures is true for a user


OAGr:
Just hide the link in the UI

PR #580, opened 2019-08-26
Hides Leaderboard tab

RomanGalochkin:
#579


OAGr:
That was quick, nice work!

Issue #557, opened 2019-08-19
Community AgentMeasurable score tables


OAGr:
Communities should now have a "By Question" leaderboard tab.

![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/03796aef-5445-4e96-92d0-58ac30951e3e)

This should have the following columns:
- Question (link)
- Agent (link)
- Total Score (score)
- Prediction Count
- Resolution Time (in commonly-used time format)

PR #585, opened 2019-08-27
Feature/557 - leaderboard by question

RomanGalochkin:
#557 

Please, consider the PR after all previous are merged.


OAGr:
Overall, looks good. There are some issues though (sorry if this wasn't obvious earlier)

1. Items should only show up on the "By Question" list if the agent made at least one prediction for that measurable, **and** that measurable was resolved. Otherwise, there's no score associated with it (yet).

2. Measurable-agent pairs should be unique. I'm seeing a lot of duplication for some reason. The pair exists to include all information about that grouping. 

image

RomanGalochkin:
Ok, sure, will fix in 5 minutes

RomanGalochkin:
Fixed.

RomanGalochkin:
I feel that we should count Measurements if they are Competitive, Objective, Aggregation but not Comment? (predictionCountTotal)

OAGr:
Hm... we should probably just count competitive + aggregation measurements.

RomanGalochkin:
ok, will update

OAGr:
Hm... just do in a different PR or something, want to get these other stories in.

RomanGalochkin:
sure

PR #586, opened 2019-08-28
Feature/559

RomanGalochkin:
Please, change the direction of PR on "master" branch. It was done to show code changes.

### DIFF
--- a/packages/server/src/models/agent-channel.js
+++ b/packages/server/src/models/agent-channel.js
@@ -1,0 +1,41 @@
+module.exports = (sequelize, DataTypes) => {
+  const AgentChannel = sequelize.define('AgentChannel', {
+    id: {
+      type: DataTypes.UUID(),
+      primaryKey: true,
+      defaultValue: sequelize.fn('uuid_generate_v4'),
+      allowNull: false,
+    },
+    agentId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    channelId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    primaryPointScore: {
+      type: DataTypes.FLOAT,
+      allowNull: false,
+    },
+    createdAt: {
+      type: DataTypes.DATE,
+      defaultValue: DataTypes.NOW,
+    },
+    updatedAt: {
+      type: DataTypes.DATE,
+      defaultValue: DataTypes.NOW,
+    },
+  });
+
+  AgentChannel.associate = function associate(models) {
+    AgentChannel.Agent = AgentChannel.belongsTo(models.Agent, {
+      foreignKey: 'agentId',
+    });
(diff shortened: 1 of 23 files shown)


packages/server/src/models/agent-channel.js
```diff
@@ -0,0 +1,49 @@
...
+      allowNull: false,
+    },
+    numberOfPredictions: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+    },
+    numberOfQuestionsScored: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+    },
```
OAGr: Is it important that ``numberOfPredictions`` and ``numberOfQuestionsScored`` are stored in the db? They seem to be caches. ``primaryPointScore`` too, though that is a bit more computation.

OAGr: If we do store them in the db, maybe we could add ``cache`` to the name to make it obvious what they are doing? Like, ``numberOfPredictionCache``

OAGr: Also, how do these get updated as they change? Is there logic for that yet? 

How about instead of doing it this way, you just calculate these on demand? I don't think we need to allow users to filter for these in the graphQL layer.

packages/server/src/schema.js
```diff
@@ -107,6 +107,17 @@ const schema = new graphql.GraphQLSchema({
         resolve: resolvers.agentMeasurables.all,
       },
 
+      agentChannels: {
+        type: types.agentChannels.agentChannelsConnection,
+        args: {
+          ...types.common.connectionArguments,
+          minNumberOfPredictions: { type: graphql.GraphQLInt },
```
OAGr: So the use case here is to filter by the minNumberOfPredictions? Why is this needed? I could imagine it being useful later on, not sure about now.

RomanGalochkin: There is no huge need to write logic with redundant conditions, instead of this we are able to give to the clients (UI) possibilities to solve what and how they want to get. It is not pattern, but I believes it is more flexible then hard code.

OAGr: ok

packages/server/src/models/agent-channel.js
RomanGalochkin: 1. Since we all are not sure about future code changes here we use "views" instead of "tables". "Views" are quite the same SQL queries, but they are already saved in the DB to make SQL queries right to them as well as to ordinary tables (for instance, "SELECT * FROM ViewA"). 
2. Hence, "views" are dynamic. Here no need to update anything. 
3. To get counters from DB is faster then make another DB queries.

OAGr: Oh nice, makes sense. I didn't notice that you were using Views.

PR #587, opened 2019-08-28
Feature/560 - leaderboard by member

RomanGalochkin:
Please, change the base branch on "master" after all previous PRs are merged. Have done it to show code changes.

### DIFF
--- a/packages/client/src/components/LeaderboardTable.re
+++ b/packages/client/src/components/LeaderboardTable.re
@@ -81,7 +81,7 @@ module Columns = {
           switch (r.predictionCountTotal) {
           | Some(predictionCountTotal) =>
             predictionCountTotal |> Js.Int.toString |> Utils.ste
-          | _ => "0" |> Utils.ste
+          | _ => "" |> Utils.ste
           },
       ~flex=1,
       (),
--- a/packages/client/src/components/LeaderboardTable.re
+++ b/packages/client/src/components/LeaderboardTable.re
@@ -87,6 +87,20 @@ module Columns = {
       (),
     );
 
+  let predictedMeasurablesCount =
+    Table.Column.make(
+      ~name="Predicted Question Count" |> Utils.ste,
+      ~render=
+        (r: record) =>
+          switch (r.numberOfQuestionsScored) {
+          | Some(numberOfQuestionsScored) =>
+            numberOfQuestionsScored |> string_of_int |> Utils.ste
+          | _ => "0" |> Utils.ste
+          },
+      ~flex=1,
+      (),
+    );
+
   let time =
     Table.Column.make(
       ~name="Time" |> Utils.ste,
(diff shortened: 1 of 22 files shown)


packages/client/src/components/LeaderboardTable.re
```diff
@@ -53,7 +53,7 @@ module Columns = {
       ~render=
         (r: record) =>
           switch (r.pointScore) {
-          | Some(pointScore) => pointScore |> string_of_float |> Utils.ste
+          | Some(pointScore) => pointScore |> Js.Float.toString |> Utils.ste
```
OAGr: why? I prefer the E. library, so this could be standardized.

RomanGalochkin: string_of_float is deprecated

OAGr: Ah, can you put the functionality in the E library then, so all of our library use is one place, especially for ``Float``? Like, ``E.Float`` or whatever it is.

RomanGalochkin: yes, thanks

packages/client/src/components/LeaderboardTable.re
```diff
@@ -80,7 +80,21 @@ module Columns = {
         (r: record) =>
           switch (r.predictionCountTotal) {
           | Some(predictionCountTotal) =>
-            predictionCountTotal |> string_of_int |> Utils.ste
+            predictionCountTotal |> Js.Int.toString |> Utils.ste
+          | _ => "0" |> Utils.ste
```
OAGr: This doesn't seem right to me. If the predictionCountTotal is empty, we don't know if it's zero. I'd prefer just showing an empty string or a null react element.

packages/client/src/graphql/AgentChannelsGet.re
```diff
@@ -0,0 +1,149 @@
...
+      ~fn=
+        Query.make(
+          ~channelId?,
+          ~minNumberOfPredictions?,
+          ~minNumberOfQuestionsScored?,
+        ),
+      (),
+    );
+  componentMaker(query, innerComponentFn);
+};
```
OAGr: Just curious, how long does it take you to write these files? They're kinda long and hypothetically we may be able to do something else to not need them to be so long later on.

RomanGalochkin: Mmm, it was fast, so there is no hard need to make scaffolding or similar at the moment.

OAGr: Cool, good work

packages/client/src/components/LeaderboardTable.re
RomanGalochkin: Fixed.

packages/client/src/components/LeaderboardTable.re
RomanGalochkin: Fixed.

RomanGalochkin:
Updated.

OAGr:
https://github.com/foretold-app/foretold/pull/589

PR #588, opened 2019-08-28
Feature/581

RomanGalochkin:


OAGr:
This looks good, but I did make a mistake writing it apparently (sorry!) Can you remove the "prediction" and "recent aggregate" columns (in any way you want to). I don't think they will be needed in the future; it's a bit complicated, as it's not obvious which prediction is chosen; and the last/first aren't particularly important for scoring.

RomanGalochkin:
Have updated this PR. Now that table looks like:

Screenshot_20190829_100017
Screenshot_20190829_100030

OAGr:
Looks good, thanks!

PR #615, opened 2019-08-30
Improvements/601

RomanGalochkin:


RomanGalochkin:
Screenshot_20190830_163007

RomanGalochkin:
Screenshot_20190830_163031

RomanGalochkin:
How do you think it is enough for the first pass?

OAGr:
yea, that's fine. If the code is done, submit it, then we could make more passes later.

PR #627, opened 2019-09-02
Feature/626

RomanGalochkin:
Just for testing. It seems it works.


OAGr:
Hm... I think this is getting there, but has more work to go.

Let's make this a bit simpler:
1) Just fetch all aggregates, rather than the future ones, and the previous one.
2) Instead of this returning multiple results, one for each prediction, and each with the ``recentResult`` (which will always be the same), we should just have one larger result.

The interface should return one object like the following:
```js
{
 allAggregations: [{id, createdAt, value, type}],
 recentResult: [{id, createdAt, value, type}],
 predictions: [{id, createdAt, value, type}],
 agentId,
 measurableId,
 measurableCreatedAt
 (maybe the agentMeasurable Id too)
}

We'll be processing all predictions at once, so just need one query for all of them.
```

image

RomanGalochkin:
Ok

RomanGalochkin:
Will reopen later

RomanGalochkin:
Do not understand what exactly should I do here

RomanGalochkin:
It is not a big deal to get measurements, but what is the next step?

RomanGalochkin:
The solution is fully depended of the way it will be used on the next steps

RomanGalochkin:
I have updated this a bit, but feel that it works some.

OAGr:
I think this is good for this purpose, will play with it more later. Thanks!

PR #630, opened 2019-09-03
Fixes/619

RomanGalochkin:


RomanGalochkin:
And here I've fixed the long previsions of score points on "Prediction Scores" tab.

PR #636, opened 2019-09-04
Improvements/624

RomanGalochkin:
I think it will be similar to this. Step by step we add functionality into this lib.


OAGr:
Why did you delete ``packages/components/package.json``? We still need that.

packages/measurement-value/float-cdf.js
```diff
@@ -0,0 +1,42 @@
+const _ = require('lodash');
+
+const { MeasurementValue } = require('./measurement-value');
+
+class FloatCdf extends MeasurementValue {
+  xs(v) {
+    return v === undefined
+      ? _.get(this.input, 'xs')
+      : _.set(this.input, 'xs', v);
```
OAGr: can you use an es6 getter/setter instead?
https://stackoverflow.com/questions/28222276/what-are-getters-and-setters-for-in-ecmascript-6-classes

packages/measurement-value/float-cdf.js
```diff
@@ -0,0 +1,42 @@
...
+
+  sizeYs() {
+    return _.size(this.ys());
+  }
+
+  validate() {
+    if (this.sizeXs() !== this.sizeYs()) {
+      throw new Error('Xs and Ys should be the same size.');
+    }
+    if (this.sizeXs() > FloatCdf.MAX_XS) {
```
OAGr: This should be a server thing, maybe configurable here, but not forced in this lib.

packages/measurement-value/float-cdf.js
```diff
@@ -0,0 +1,42 @@
+const _ = require('lodash');
+
+const { MeasurementValue } = require('./measurement-value');
```
OAGr: what would you think of this lib importing the ``cdf`` lib, then this class using the ``Cdf`` class underneath?

RomanGalochkin:
> Why did you delete `packages/components/package.json`? We still need that.

Thanks, will recover.

packages/measurement-value/float-cdf.js
RomanGalochkin: I do not like setter/getters according to my own experience and knowledge they mislead clients of classes. We should write either "getXs()" or "xs()". But do not get properties directly. May this book shows it (https://www.amazon.com/Code-Complete-Practical-Handbook-Construction/dp/0735619670) may be another.

RomanGalochkin: There are exceptions but not in this case.

OAGr: Ok, I'm find with that.

Actually, now that I think about it, why do we even need a setter? Can't we just set these when we create a new instance. I assume they really shouldn't need to be mutated.

RomanGalochkin: Ok, will remove setter.

packages/measurement-value/float-cdf.js
RomanGalochkin: I propose to forget about this for now, later we can fix it. Here is no much sense to write configuration providers and so on.

OAGr: Ok, sure

packages/measurement-value/float-cdf.js
RomanGalochkin: It depends of code usage cases, but I believe that yes we can do it here.

RomanGalochkin:
Updated.

OAGr:
I think this is a good outline. We can add more stories to flesh it out soon/later.

PR #651, opened 2019-09-09
Fixes/650

RomanGalochkin:
Here are some math questions, in "distributionInputDistributionOutputMarketless" sometimes we are going to calculate log2 from negative number.

### DIFF
--- a/packages/cdf/lib/continuousDistribution.js
+++ b/packages/cdf/lib/continuousDistribution.js
@@ -7,6 +7,12 @@ class ContinuousDistribution {
    * @param {number[]} ys
    */
   constructor(xs, ys) {
+    if (!_.isArray(xs)) {
+      throw new Error('XS should be an array.');
+    }
+    if (!_.isArray(ys)) {
+      throw new Error('YS should be an array.');
+    }
     if (!this.validateHasLength(xs)) {
       throw new Error('You need at least one element.');
     }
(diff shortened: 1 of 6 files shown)


RomanGalochkin:
To debug this on your local machine 
1. Get DB url like this
heroku config:get DATABASE_URL -a prediction-backend
2. Put this string into docker-compose.override.yml in "environment" section
```
version: "2"

services:

  app:
    volumes:
      - ./packages:/opt/app/packages
    environment:
      - DATABASE_URL=postgres://
```
3. Run `docker-compose up -d app` to update application settings.
4. In "./foretold/packages/server/config/config.json" set this
```
{
  "development": {
    "use_env_variable": "DATABASE_URL",
    "dialect": "postgres",
    "dialectOptions": {
      "ssl": true
    }
  },
  "production": {
    "use_env_variable": "DATABASE_URL",
    "dialect": "postgres"
  }
}
```
5. Now your machine is connected with remote prod DB (on Heroku). After work do not forget to recover config files.

OAGr:
Hm... I'll mess with it, thanks.

OAGr:
> "Here are some math questions, in "distributionInputDistributionOutputMarketless" sometimes we are going to calculate log2 from negative number."

-> The params for log2 should really never be negative. This should be fixed, maybe separately. I'll investigate.

PR #659, opened 2019-09-11
Improvements/317 2

RomanGalochkin:
- Measurable Index bug has been fixed;
- FC lib has been synchronized;


packages/client/src/FC/Base/FC__Button.re
```diff
@@ -77,39 +89,38 @@ let styles =
...
+      ~size=Large,
       ~isDisabled=false,
       ~fullWidth=false,
-      ~verticalPadding=`px(4),
       ~className="",
       children,
     ) => {
   ...component,
   render: _self =>
-    <FC__Link
```
OAGr: I thought the FC__Link had some fancy-ish href properties. Are any of those needed?

packages/client/src/FC/Base/FC__Globals.re
```diff
@@ -1,36 +1,24 @@
...
-        ),
+        lineHeight(`em(1.5)),
+      ],
+    ),
+    global(
+      "button",
+      [
+        fontFamily(FC__Settings.Text.standardFont),
+        fontSize(`px(16)),
+        lineHeight(`em(1.5)),
```
OAGr: Why are these being set here? Can we set them in in the button css instead? Having global properties is a bit of an anti-pattern, unless really needed.

packages/client/src/FC/Base/FC__Button.re
RomanGalochkin: Sorry, did not get properly.

packages/client/src/FC/Base/FC__Globals.re
RomanGalochkin: The main client of the app has global settings which are imported with Antd lib, so this is why it had been placed here. Will change, thanks.

packages/client/src/pages/measurables/MeasurableIndex.re
```diff
@@ -1,43 +1,36 @@
...
+              MeasurableIndex__Components.toLayoutInput(
+                reducerParams.send,
+                searchParams,
+                statsQuery,
+                MeasurableIndex__Logic.make({
+                  reducerParams,
+                  loggedInUser,
+                  channelQuery,
+                  seriesQuery,
+                }),
```
OAGr: I was kind of proud of the ``load3Queries`` fn, but this is more compact. Maybe with hooks we should eventually be able to clean this up even more.

RomanGalochkin: It was not readable, sorry, we should first of all write code for humans. Only after this refactoring I found a way to fix the issue.

packages/components/src/Charts/FC__CdfChart__Small.re
```diff
@@ -31,8 +37,8 @@ let make =
       <FC__CdfChart__Base
         width=200
         height=40
-        minX
-        maxX
+        ?minX
+        ?maxX
```
OAGr: +1 for making these optional.

packages/client/src/FC/Base/FC__Button.re
OAGr: Never mind; I thought it did some things it doesn't actually do.

packages/client/src/pages/measurables/MeasurableIndex.re
OAGr: sg

OAGr:
I played with this locally, seems to work well!

RomanGalochkin:
Updated, I have moved buttons styles in button component.

PR #664, opened 2019-09-13
New PredictionAnalysis library

OAGr:
Right now this does a few things. The main bits are integrated into the ``server`` lib, to replace the existing prediction scoring, and add a first set of question scoring.

This still needs:
- Tests to work
- Possibly removing Babel from the js lib, or upgrading it to es6.
- Maybe a ReadMe.
- Additional cleanup/improvement, but some of that could come in future PRs.


RomanGalochkin:
Here I have fixed code style and some issues with file names through a pair of commits. It  seems like you use Windows or similar where is no difference in files names cases.

OAGr:
@RomanGalochkin The tests are still breaking in Travis. Can you either try to fix it here, or make a new story for them?

OAGr:
Btw, I use a mac with VSCode

RomanGalochkin:
Ok, will figure it out in some minutes

OAGr:
Thanks!

PR #687, opened 2019-09-19
Feature/680

RomanGalochkin:
It turns out that to make ordering through a score point it is not most obvious thing. After a discussion within our team we have combined approaches in this google doc:
https://docs.google.com/document/d/1bEHhT882NDqU17J9a-pkdlasIxdagqcF1fqMwiitJCE/edit?usp=sharing


OAGr:
Looks good. Sorting isn't that important now, we could consider it more later.

PR #695, opened 2019-09-20
Adds a bundler

RomanGalochkin:


OAGr:
Does this make it work in the observable debugger? Can you try testing it on that? Do you need my access for npm for publishing or something?

RomanGalochkin:
I have copied the bundle file on our server and it works. So what should I do here else? Should I set up something? Do not follow.

OAGr:
I'll try this and see.

PR #701, opened 2019-09-24
Refactoring, freezes versions and removes "package.lock" 

RomanGalochkin:
Then bigger the system is becoming then more complex the package management and development are. I have spent much time to understand how to get it up working. After ordinary update all has gone down in development. 

1. So, I propose "to freeze" versions. 
2. Use only YARN. That means "package.lock" files are obsolete.

It will prevent future dependencies issues and will help to improve the setting of the system.


RomanGalochkin:
To maintain versions did you mean?

OAGr:
I was thinking that ``lerna bootstrap`` would ideally solve the dependency issues without needing to freeze versions. 

I'm wondering if there are other stories we should add to ensure that all packages work well together. For instance, Travis should probably test all packages, to be safe.

RomanGalochkin:
I think we should freeze the most of packages but not our own. For instance we could use @foretold/cdf: ^1.0.9, but not for another like "rationale" etc.

RomanGalochkin:
I am for a predictable stage of the application. But it is up to you sure.

PR #706, opened 2019-09-25
Improvements/686 remove babel

RomanGalochkin:
Yes, we have duplicated packages. For instance "antd" within "client" and "antd" within a "bs-ant-design" library. The "bs-ant-design" library has "andt@3.8.2-3.15.0" dependency, "client" has dependency of "antd@^3.17.0". So to remove a duplicate the "bs-ant-design" should have the upper version of "antd". If we had removed the "antd" lib we would have saved 1MB.

Between I have install "parcel-plugin-bundle-visualiser". After "yarn clietn/build" it creates "/report.html" file. To see this file run "cd ./packages/client && export PORT=1234 && yarn serve" after what run the explorer "localhost:1234/report.html".

Screenshot_20190925_153044


OAGr:
Thanks!
Curious, how much does this decrease the total file size?

RomanGalochkin:
-1MB

RomanGalochkin:
Sorry, did not get your question right, this PR does not effect the total size at all.

PR #720, opened 2019-10-02
Improvements/443

RomanGalochkin:
1. Removed the "client/lib/ken" folder.
2. Fixed names of measurables with "ken js".


packages/ken-js/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
OAGr: why do we need this file, instead of using bs-ken?

RomanGalochkin: Yes, sure, give me a pair of miniutes

OAGr:
Kudos for drying up the Ken integration; using the bs-ken lib instead of having the files copied. Will merge and see how staging deals with it.

PR #740, opened 2019-10-08
Feature/712

RomanGalochkin:


OAGr:
I'm curious; isn't it the client that would not use SSL? Why is this a fix on the server side? The client codebase always sends messages to the server via SSL, I believe.

PR #744, opened 2019-10-09
Fixes too much redirecting

RomanGalochkin:


RomanGalochkin:
Here I have added "netlify.toml" file. I do not know will it work or no. We need to test it. But I do not have an access to publishing "the client part" to "Netlify.com". Please, try it and let me know after this.

Between for the future we need either to set "netlify.com" and for "staging" too or use only "heroku.com". The staging and production environment have to be similar.

RomanGalochkin:
My last comment is not actual any more since "Foretold" does not use "Netlify.com".

RomanGalochkin:
I think it is ready now, have tested with on staging version.

PR #748, opened 2019-10-09
Improvements/735

RomanGalochkin:
The first issue is not solved since I do not understand how to pass an error to the client in the production mode. It is weird. How do other companies work with it. Do not understand. 

One of the solutions is may be to use another interface like "query validationInput" to get a list of possible errors and warnings before doing a request.


packages/client/src/layout/Navigator.re
```diff
@@ -14,7 +14,10 @@ let make = (~route: Route.t, ~loggedInUser: option(Types.user), _children) => {
     | (AgentIndex, _) => AgentIndex'.toEl(loggedInUser)
     | (EntityShow(id), _) => EntityShow'.toEl({id: id}, loggedInUser)
     | (EntityIndex, _) => EntityIndex'.toEl(loggedInUser)
-    | (Profile, _) => Profile'.toEl(loggedInUser)
+    | (Profile, Some(loggedInUser)) =>
+      <FillWithSidebar loggedInUser>
+        <Profile loggedInUser />
+      </FillWithSidebar>
```
OAGr: Why was this needed? Seems a bit hacky.

RomanGalochkin: I have to use <ComponentStyle> it is more readable and understandable. If we use functions instead of components we will make the hierarchy of the app too complicated. How it is now for instance with the "authentication" flow.

OAGr:
Ok, we'll go with this for now, will need some way of getting errors to the client. I could give it a go.

packages/client/src/layout/Navigator.re
OAGr: Ok, that works. I'm not sure, will keep in mind.

PR #757, opened 2019-10-11
Adds intercom js

RomanGalochkin:


packages/client/src/intercom.js
```diff
@@ -0,0 +1,46 @@
...
+        w.addEventListener('load', l, false);
+      }
+    }
+  })();
+
+  window.Intercom('boot', {
+    app_id: 'gx67sx4f',
+    name: 'Foretold', // Full name
+    email: 'internal-tech@foretold.io', // Email address
+    created_at: Math.floor(Date.now() / 1000),
```
OAGr: I think this should be the user's name/email/created at?

RomanGalochkin: Oh, sorry, sure, I did understand it from the first pass. Will change.

RomanGalochkin:
I think it is ready.

packages/client/src/lib/intercom.js
```diff
@@ -0,0 +1,41 @@
...
+      };
+      l();
+    }
+  })();
+
+  var options = {
+    app_id: 'gx67sx4f',
+    name: name,
+    email: email,
+    created_at: Math.floor(Date.now() / 1000),
```
OAGr: I'm quite sure created_at is supposed to be when the user was created, not the current date.

packages/client/src/lib/intercom.js
```diff
@@ -0,0 +1,41 @@
...
+        s.src = 'https://widget.intercom.io/widget/gx67sx4f';
+        var x = d.getElementsByTagName('script')[0];
+        x.parentNode.insertBefore(s, x);
+      };
+      l();
+    }
+  })();
+
+  var options = {
+    app_id: 'gx67sx4f',
```
OAGr: Can this be stored in a config file or similar, as opposed to being a magic variable here?

RomanGalochkin: Ok.

packages/client/src/lib/intercom.js
RomanGalochkin: Yes, will fix.

packages/client/src/utils/Env.re
```diff
@@ -1,17 +1,22 @@
...
 [@bs.val] external auth0ClientIdEnv: string = "window.ENV.AUTH0_CLIENT_ID";
+[@bs.val] external modeEnv: string = "window.ENV.MODE";
 [@bs.val] external origin: string = "window.location.origin";
 
 let auth0Domain = auth0DomainEnv;
 let auth0ClientId = auth0ClientIdEnv;
 let redirectUrl = origin ++ "/callback";
 let logoutUrl = origin ++ "/login";
 let serverUrl = apiUrlEnv;
+let mode = modeEnv;
```
OAGr: what is "MODE"? Maybe "CLIENT_ENV" or "JS_ENV" or something?

RomanGalochkin: Think that "CLIENT_ENV" is better.

packages/client/src/utils/Env.re
```diff
@@ -1,17 +1,22 @@
...
+let prod = mode == "production";
+let dev = mode == "development";
```
OAGr: I think this could be done better. Feels like it should be a type, like,
``type appEnvironment = Development | Production | Error``

OAGr:
Great, I look forward to using it.

PR #764, opened 2019-10-14
Feature/733 check session

RomanGalochkin:


packages/client/src/graphql/users/UserGet.re
```diff
@@ -96,7 +96,7 @@ module Query = [%graphql
 module QueryComponent = ReasonApollo.CreateQuery(Query);
 
 let inner = innerComponentFn => {
-  <QueryComponent>
+  <QueryComponent pollInterval={60 * 1000}>
```
OAGr: Does this need to poll, even if their email has been validated? I wouldn't expect the user model to update much otherwise.

packages/client/src/components/CheckSession.re
```diff
@@ -0,0 +1,46 @@
+let component = ReasonReact.statelessComponent("CheckSession");
+
+let make = _children => {
+  ...component,
+  render: _self => {
+    <Providers.AppContext.Consumer>
+      ...{({loggedInUser}) =>
+        switch (loggedInUser) {
+        | Some(loggedInUser) =>
+          <UserAccessTokenUpdate.Mutation>
```
OAGr: Is this running this mutation for all users, even if their email addresses and everything are already valid?

OAGr:
Didn't mean to close

packages/client/src/graphql/users/UserGet.re
RomanGalochkin: No, it does not need to poll if user has been validated. It was the simplest form of updating user info. I sure there are another ways.

packages/client/src/components/CheckSession.re
RomanGalochkin: Yes. And for what I am going to give an explanation below.

RomanGalochkin:
The main purpose of this PR is to update user info from "Auth0.com" service and then update this info on the client side. But there are some another issues.

1. We use "Auth0.com" service.
2. This service gives us two types of tokens. The first one is JWT with user info and the second one is an "Access Token". Only the last one we can use to communicate with "Auth0.com" to get fresh user info.
3. An "Access Token" has an expiration time. The current value is 2 hours, however the maximum value is 1 month (it could be set in "Auth0.com" control panel).
4. So to make queries to "Auth0.com" for user info we need to store an "Access Token".
5. Since this token has expiration time we need to refresh this token on the client side.

I do not like additional conditions in data flows and logic flows. This is why I have split this on three main not depended flows:
1. The first flow is updating users' "Access Tokens". It runs "web-auth.checkSession" functions on the client side, takes fresh "Access Token" and send this to the server side.
2. The second flow is working on the server side. There is new cron-job. For instance it is run every minute. It takes users without verified emails and with an "Access Token", makes query to "Auth0.com" and updates user info.
3. The third flow is just updating user info on the client side.

It is simple to understand. It is simple to split. It is simple to correct. But sure we could optimize something for instance we could not poll user info if user has been validated their email. But in the future there will be issues with a count of "New Notifications" or similar. 

Or even the simplest form of doing it: to get a fresh "Access Token" then to make query to the server and get updated info back in a background. In this case we do not need to store "Access Token". This thoughts come to me right now when I was writing it.

OAGr:
Ok. I'll merge now and try it out in practice. Lets continue to think about this. I'm a bit frustrated by how complicated it seemed to want to be, but understand solutions like this may be necessary.

PR #768, opened 2019-10-15
Feature/593

RomanGalochkin:
It was one of important fundamental experiences I have got within this project. I mean something like two way data type determination. When both sides work with the same. The cursor now it is not just a string, and has its own abstraction the "Cursor".

In the future we should add another types for limit length of input strings.


OAGr:
Great, nice work!

PR #807, opened 2019-10-30
Feature/788

RomanGalochkin:


OAGr:
looks good to me, will merge after it passes tests.

PR #817, opened 2019-11-01
Page of notebook

RomanGalochkin:


packages/client/src/components/shared/Padding.re
```diff
@@ -0,0 +1,9 @@
+[@bs.config {jsx: 3}];
+
+module Styles = {
+  open Css;
+  let main = style([padding(`em(2.))]);
+};
+
+[@react.component]
+let make = (~children) => <> <div className=Styles.main> children </div> </>;
```
OAGr: I'm not sure I'm a fan of these files. It seems like they should be in the components module, and also I'm really not sure if this is the best way of doing it. 

Honestly, I'd probably prefer we just did this directly in the code, like
```reasonml
<div className=Css.(style([padding(`em(2.))]))> children </div> 
```

I'll merge for now, I'll mess with separately.

PR #824, opened 2019-11-04
=mm(1 to 100...). Part 1. update Function.js

NunoSempere:
Allows the program to accept inputs of the form =mm(1 to 100, normal(10,5)) by replacing the part like  [1 to 100] with lognormal(...) when appropriate

### DIFF
--- a/packages/guesstimator/formatter/formatters/Function.js
+++ b/packages/guesstimator/formatter/formatters/Function.js
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import {regexNumberToNumberIntoLognormal} from './lib'
+import {NumberToNumberIntoLognormal} from './lib'
 
 //FIXME
 export const item = {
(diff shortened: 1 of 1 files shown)


OAGr:
Kudos for getting this to work. However, looking at this, I think this could be improved.

1. First, please submit PRs in whole units when one requires the other. If I'd accept this and not the other things would break, which is sub-ideal. In cases where different PRs independently achieve different pieces of functionality they make sense, but here less so.

2.  In general, I recommend aiming for comments to above functions, rather than in the middle of them. This keeps things cleaner. 

3. I'm not a fan of the name ``NumberToNumberIntoLognormal``. Names are tricky. This name made me think it was converting a number into another number and that into lognormal, which would be weird. We already refer to this pattern in the codebase now, as ``DistributionTextUpTo``; I'd hope for something similar to that for this. Also, I would prefer it to be lowercase. In general lowercase is preferred for functions, though this specific codebase does have some exceptions, which is kind of messy and semi-random. For what this is doing, I could imagine calling at least one function here, ``numberShorthandCompilationStep`` or similar. 

4. I don't love the fact that the ``replacerNumberToNumber`` functionality is so similar to the existing code in ``DistributionLognormal.js``. This should be dried up. One way to do that is something like having a ``lib`` directory on the same level as the ``formatter`` and ``sampler`` directory, with a file called ``distributionMath.js`` or similar, and have the main bit of this in there. Note that one thing that should really happen later (after this PR) is that the ``samplers`` and ``formatter`` libs should be wrapped in one ``src`` folder or similar; what is happening now is kinda sloppy. Similarly, there already is a regex for "3 to 10" and other useful params, that you can see in ``DistributionTextUpTo.js``; I'm not sure how to best use this but imagine it's possible.

5. Some things this code doesn't do (I believe) is:
- Handle existing supported syntaxes like "3-10"
- Handle the suffixes, like "3K to 10K"
- Handle cases where the low number is 0 or less, like, "-3K to 3K".

These really should be handled to match user expectations. These are already done in the app, the trick is figuring out how to either reuse that code or move it to other places so that it can be reused.

6. I think this is pretty messy. I think some tests would be pretty useful, to test the ``NumberToNumberIntoLognormal`` fn specifically.

PR #827, opened 2019-11-04
Shorthand formatting step

NunoSempere:
Addressed the cosmetic changes:
1. PRs as a whole unit
2. Write the comments above the functions
3. Renamed NumberToNumberIntoLognormal to shorthandIntoLognormalFormattingStep, which might be more to your liking.

Did not address:
4. Similarities to DistributionLognormal.js. This sounds like something which would take some more thought.
6. Tests. This seems important. Will do next.

Re 5, Special cases. I did some testing:
5a. This doesn't address other syntaxes, like "3-10" (or "3 - 10", or whitespaces). Which other syntaxes are there? I haven't seen people use them in foretold, but I'm happy to add them. Also, the input "3-10" doesn't produce any distribution in foretold right now.
5b. I am uncertain about whether this addresses things like "3K to 4K". Right now, foretold.io doesn't accept =lognormal(3.5K, 10) as an input, so it probably doesn't.
5c. This doesn't handle cases where the low number is 0 or less, like, "-3K to 3K". Right now, "=lognormal(-10, 2)" also breaks foretold. In fact, it is accepted as an input, but the distribution produced is a different one. Same with the inputs: "-3K to 3K", "-10 to 1".

tl;dr: The special cases are mostly not supported on foretold right now, or break it. I haven't seen people use them.

### DIFF
--- a/packages/guesstimator/formatter/formatters/Function.js
+++ b/packages/guesstimator/formatter/formatters/Function.js
@@ -1,8 +1,15 @@
 import _ from "lodash";
+import {shorthandIntoLognormalFormattingStep} from './lib'
+
 //FIXME
+
+// The function: shorthandIntoLognormalFormattingStep
+// transforms strings like "=mm(normal(10,5), 1 to 100))
+// into strings like "=mm(normal(10,5), lognormal(50.1, 1.4))
+
 export const item = {
   formatterName: 'FUNCTION',
   matches({text}) { return !!text && text.startsWith('=') },
   error({text}) { return !_.isEmpty(text.slice(1)) ? {} : {type: 1, subType: 2}},
-  format({text}) { return {guesstimateType: 'FUNCTION', text: text.slice(1)} },
+  format({text}) { return {guesstimateType: 'FUNCTION', text: (shorthandIntoLognormalFormattingStep(text)).slice(1)} },
 }
--- a/packages/guesstimator/formatter/formatters/Function.js
+++ b/packages/guesstimator/formatter/formatters/Function.js
@@ -1,12 +1,6 @@
 import _ from "lodash";
 import {shorthandIntoLognormalFormattingStep} from './lib'
 
-//FIXME
-
-// The function: shorthandIntoLognormalFormattingStep
-// transforms strings like "=mm(normal(10,5), 1 to 100))
-// into strings like "=mm(normal(10,5), lognormal(50.1, 1.4))
-
 export const item = {
   formatterName: 'FUNCTION',
   matches({text}) { return !!text && text.startsWith('=') },
--- a/packages/guesstimator/formatter/formatters/lib.js
+++ b/packages/guesstimator/formatter/formatters/lib.js
@@ -71,8 +71,9 @@ export function shorthandIntoLognormalFormattingStep(text){
     
         p1 = Number(p1)
         p2 = Number(p2)
-        let logHigh = math.log(p1)
-        let logLow = math.log(p2)
+        
+        let logLow = math.log(p1)
+        let logHigh = math.log(p2)
 
         let mean = math.mean(p1,p2)
         let stdev = (logHigh-logLow) / (2*1.645)
--- a/packages/guesstimator/formatter/formatters/lib.js
+++ b/packages/guesstimator/formatter/formatters/lib.js
@@ -89,7 +89,7 @@ export function shorthandIntoLognormalFormattingStep(text){
 
     let regNumberToNumber= new RegExp("([0-9]+) to ([0-9]+)", "g");
     
-    return text.replace(regNumberToNumber, shorthandIntoLognormalReplacerStep);
+    return text.replace(regNumberToNumber, shorthandIntoLognormalReplacer);
 
 }
 
--- a/packages/guesstimator/formatter/formatters/lib.js
+++ b/packages/guesstimator/formatter/formatters/lib.js
@@ -1,5 +1,6 @@
 import _ from "lodash";
 import math from 'mathjs';  
+import {distributionUpToIntoLognormal} from "../../lib/distributionMath.js";
 
 const SUFFIXES = {
   '%': -2,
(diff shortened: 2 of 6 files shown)


packages/guesstimator/formatter/formatters/lib.js
```diff
@@ -57,3 +57,39 @@ export function regexBasedFormatter(re, guesstimateTypeFn = getGuesstimateType,
     _numbers(text) { return _.chunk(text.match(re).slice(1), 2).map(([num, suffix]) => parseNumber(num, suffix)) },
   }
 }
+
+
+
+
+
```
OAGr: 1 newline is fine, 5 is unnecessary and unusual for the current codebase.

packages/guesstimator/formatter/formatters/lib.js
```diff
@@ -57,3 +57,39 @@ export function regexBasedFormatter(re, guesstimateTypeFn = getGuesstimateType,
...
+// The function: shorthandIntoLognormalFormattingStep 
+// is used in foretold/packages/guesstimator/formatter/formatters/Function.js
```
OAGr: I'm not a fan of comments like this, because:
1. It's easy enough to search the codebase to figure out where it's used.
2. The presence of this implies that if someone were to use this elsewhere, they should also need to add to the comment. This would get pretty messy pretty quickly, and likely get out of sync.

packages/guesstimator/formatter/formatters/lib.js
```diff
@@ -57,3 +57,39 @@ export function regexBasedFormatter(re, guesstimateTypeFn = getGuesstimateType,
...
+// In order to be able to format inputs like "1 to 100" when inside a multimodal, like "=mm(normal(10,5), 1 to 100))".
+
+// The function: shorthandIntoLognormalReplacer
+// is inside shorthandIntoLognormalFormatting. This might be marginally slower, but it makes clear that it is not used by any other function.
```
OAGr: I think this is fairly obvious and doesn't need to be explicitly stated. I find semi-obvious comments hurt readability, because they make people less likely to see the useful comments.

packages/guesstimator/formatter/formatters/Function.js
```diff
@@ -1,8 +1,15 @@
 import _ from "lodash";
+import {shorthandIntoLognormalFormattingStep} from './lib'
+
 //FIXME
+
+// The function: shorthandIntoLognormalFormattingStep
+// transforms strings like "=mm(normal(10,5), 1 to 100))
+// into strings like "=mm(normal(10,5), lognormal(50.1, 1.4))
```
OAGr: This to me is a fairly useful comment, as it makes the interface straightforward. However, I think it should be in the code for shorthandIntoLognormalFormattingStep, not here, where it's used. It could be used in several places later.

packages/guesstimator/formatter/formatters/lib.js
```diff
@@ -57,3 +57,39 @@ export function regexBasedFormatter(re, guesstimateTypeFn = getGuesstimateType,
...
   }
 }
+
+
+
+
+
+// The function: shorthandIntoLognormalFormattingStep 
+// is used in foretold/packages/guesstimator/formatter/formatters/Function.js
+// In order to be able to format inputs like "1 to 100" when inside a multimodal, like "=mm(normal(10,5), 1 to 100))".
```
OAGr: I think this is the main useful line to me here. ``"a string".replace can take a function,`` is walking the line, but should be easy enough for someone to Google.

OAGr:
It seems like it's getting better, still some comments.

**On Syntaxes**
People have used the other syntaxes in Guesstimate. It looks like the supported syntaxes are [3,10], 3->10, 3:10. These are shown in the tests in formatter/formatters/tests. I feel like we should be consistent, but maybe the correct course is to remove these from Foretold (though this would break some interoperability with Guesstimate). These were originally added to Guesstimate before the "x to y" syntax was. I feel pretty awkward about this, am find holding off for now if it's a lot of work to integrate. (I really don't like non-dry code for things like this though, which is my bigger issue).

**=lognormal(3.5K, 10)**
Foretold doesn't accept this, but it does accept "3.5K to 10K". This is pretty important for larger numbers. That said, it probably should accept ``=lognormal(3.5K, 10)`` eventually, this seems like a good future addition to me. One option could be to delay handling numbers like this for a larger refactor that would also handle them in cases like ``=lognormal(3.5K, 10)``.

**=lognormal(-10, 2)**
This "breaks" foretold in the sense that the smoothing is wrong. -3K to 3K", "-10 to 1" both work fine for me. Make sure you are testing them with a question that doesn't have a min at 0 or similar.
image
image
image

NunoSempere:
4. Created a guesstimator/lib/distributionMath.js file, which feeds into both formatter and the sampler. 
// The function: distributionUpToIntoLognormal
// assumes a centered 90% confidence interval, e.g. the left endpoint
// marks 0.05% on the CDF, the right 0.95%.

Reused some of the code from formatters/lib, which fed into DistributionTextUpTo.js

5. Reusing code means that more syntaxes are handled.
6. Created a some tests.

Also, read your comments above.

packages/guesstimator/formatter/formatters/tests/lib-test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe('shorthandIntoLognormalFormattingStep', () => {
+
+  describe('Behavior:', () => {
+    const examples = [
+      ['1 to 10', 'lognormal(5.500, 0.700)'],
+      ['1K to 10K', 'lognormal(5500.000, 0.700)' ],
+      ['-1 to 10', "[unknown]"],
+      ['1->10', 'lognormal(5.500, 0.700)'],
+      ['1:10', 'lognormal(5.500, 0.700)'],
```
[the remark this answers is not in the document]
NunoSempere: Sure.

PR #834, opened 2019-11-06
Bug with session checker

RomanGalochkin:
Honestly, I did not repeat this bug but I think it will help since there was an error in the hierarchy of components.

Please, consider the previous PR firstly.


OAGr:
Hm... I'm not sure if this would fix, but happy to try.

PR #838, opened 2019-11-06
Shorthand formatting step

NunoSempere:
=mm(-10 to 10, ...) -> =mm(normal(0, 6), ...)


packages/guesstimator/formatter/formatters/lib.js
```diff
@@ -74,8 +76,9 @@ export function shorthandIntoLognormalFormattingStep(text){
       
         let low = arrayLowHigh[0]
         let high= arrayLowHigh[1]
-
-        return distributionUpToIntoLognormal(low, high)
+        
+        result = (getGuesstimateType("unknown yet", [low])=='LOGNORMAL') ? distributionUpToIntoLognormal(low, high) : distributionUpToIntoNormal(low, high)
```
OAGr: I imagine there must be some cleaner way of doing this

packages/guesstimator/lib/distributionMath.js
```diff
@@ -10,3 +10,14 @@ export function distributionUpToIntoLognormal(low, high){
...
     return `lognormal(${mean},${stdev})`
 }
+
+// The left endpoint marks 0.05% on the CDF, the right 0.95%.
+export function distributionUpToIntoNormal(low, high){
+    
+    let mean = math.mean(high, low)
+    let stdev = (high - mean) / 1.645
+    return `normal(${mean},${stdev})`
+
```
OAGr: why the spaces at the top, and especially the bottom? Also, underneath this function. Please delete.

packages/guesstimator/formatter/formatters/lib.js
NunoSempere: Yes, but not if you want to reuse code.

packages/guesstimator/lib/distributionMath.js
NunoSempere: Done

PR #839, opened 2019-11-06
Shorthand formatting step

NunoSempere:
Addressed Ozzie's last two comments.

Changed formatters/lib-test.js to avoid errors caused by empty spaces and to add the case in which =mm(-1 to 10,...) -> =mm(normal(...),...)

I also ran the tests, but had difficulties because the import statement isn't straighforwardly supported by mocha. Otherwise, I believe they work.


NunoSempere:
I realized this undid some of Ozzie's changes; please ignore.

PR #857, opened 2019-11-14
Minor Fixes

OAGr:
1. Change default resolution time to "1 month from now" on server and client.
2. Notebooks load all measurables referenced, as long as accessible.
3. Fixed bug where kde smoothing widths were 1 when max >> min, when min=0. This use case was not meant for cases where min=0.


OAGr:
This addresses:
https://github.com/foretold-app/foretold/issues/853
https://github.com/foretold-app/foretold/issues/854

PR #865, opened 2019-11-22
Minor UI updates, mostly to Sidebar

OAGr:


OAGr:
This should fix:
https://github.com/foretold-app/foretold/issues/864
https://github.com/foretold-app/foretold/issues/863

Also,
- Notebooks should support up to 128000 characters (it was 32000 before)
- Show notebook sidebars on right of entire main screen
- Increases height of comment fields (unrelated, but simple)

PR #869, opened 2019-11-27
Improvements/866 2

RomanGalochkin:


OAGr:
On the right track, but I think you misunderstood the story.

This current feature allows them to be resolved with a probability distribution (floatCdf). Instead they should be able to be resolved with a probability (dataType: percentage). This should use the percentage input that is used to predict these questions:

image


So for instance, a binary question could either be resolved like, {data: true, dataType:binary} or like, {data: 61, dataType: percentage}

RomanGalochkin:
Yes, you're right. Need to rework it.

PR #878, opened 2019-11-28
Improvements/872

RomanGalochkin:


packages/client/src/contexts/Routing.re
```diff
@@ -121,6 +121,7 @@ module Route = {
     | NotFound;
 
   let fromUrl = (url: ReasonReact.Router.url) => {
+    // @todo: Repair it, we need to move logic from here.
```
OAGr: When is this going to be repaired? Can you make and point a story for this?

PR #905, opened 2019-12-09
Bumps main dependencies

RomanGalochkin:


RomanGalochkin:
I have updated dependencies but took not the highest version of "reason-apollo-hooks" since that version does not support the old syntax of the list of queries to refetch. However, there is other issue. The refetching works either too slow either too fast that is why I used timeout in some places to give to the application a chance to update some components data. Later I will investigate this aspect and going to fix this.

PR #910, opened 2019-12-13
Update cdfChartd3.js

NunoSempere:
Creates a nicer axis.

[0,1). Up to two decimal places. 0🡒0, 0.5🡒0.5, 0.89🡒0.89
[1,1000). The number remains the same.
[1000, 1000000). To the closest K. 1000🡒1K, 2300🡒2K, etc. 
[1000000, 1000000000). To the closest million. 1000000🡒1M, 10000000🡒10M, 23M, 100M, etc.
upwards: uses other prefixes, like 1G for 1000000000, etc.

The behavior can be seen here: https://jsfiddle.net/g2mwxf8t/


OAGr:
Oh, happy to see you've worked on this!

I can't see the jsfiddle, it looks like this. Can you maybe just post a screenshot?
image

NunoSempere:
Sure, the code is:

`<script src="https://d3js.org/d3.v4.min.js"></script>

<script>

f = (d) => {
	if(Math.abs(d)<1){
  	return(d3.format(".2")(d))
  
  }else{
	  var prefix = d3.formatPrefix(".0", d)
		return(prefix(d))
  }
  
}

// alert(f(0)) // 0
// alert(f(0.5)) // 0.5
// alert(f(1)) // 1
// alert(f(10)) // 10
// alert(f(100)) // 100
// alert(f(1000)) // 1k
// alert(f(2300)) // 2k
// alert(f(10000)) // 10k
// alert(f(100000)) // 100k
// alert(f(1000000)) // 1M
// alert(f(10000000)) // 10M
</script>`

I also tried to create an observable notebook, but that was somewhat more complicated.

RomanGalochkin:
It is good that the changed file is "components/cdfChartd3.js". Because it turned out that "cdfChartd3.js" is duplicated. So, I have removed this file from "client" folder int this PR https://github.com/foretold-app/foretold/pull/907. Anyway, if there are file conflicts I will update this PR.

PR #918, opened 2019-12-16
Changed "G" to "B", and better display years

NunoSempere:
Changed 10G to 10B, (10 billions)
If the input is between 1000 and 3000, like, say, 2019, the number is rounded to the nearest integer. It would previously have been rounded to the nearest thousand. This could use some more tinkering.

### DIFF
--- a/packages/components/src/Charts/cdfChartd3.js
+++ b/packages/components/src/Charts/cdfChartd3.js
@@ -83,12 +83,23 @@ function chart() {
     xAxis
       .ticks(5)
       .tickFormat(d => {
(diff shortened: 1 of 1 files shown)


NunoSempere:
(In hindsight, I'd have continued the previous pull request, instead of creating a new one)

OAGr:
I already merged the previous PR. It's easy enough to make a new one; so I think here making a new PR was the right move.

OAGr:
I already merged the previous PR. It's easy enough to make a new one; so I think here making a new PR was the right move.

PR #923, opened 2019-12-19
Moves a measurable form into a tab

RomanGalochkin:
Sorry for marking refactoring in the same feature branch. I wanted to classify components by its types by folders. The pages components now in the "pages" folder.


OAGr:
Makes sense, I look forward to using this.

PR #925, opened 2019-12-19
Adds logger

RomanGalochkin:
First of all, we need to store logs in Google Services. Because on my local version this part of the application works as we expect.

### DIFF
--- a/packages/server/src/sync/github/app.js
+++ b/packages/server/src/sync/github/app.js
@@ -1,7 +1,10 @@
 const express = require('express');
 const bodyParser = require('body-parser');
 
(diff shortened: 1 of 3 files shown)


OAGr:
That works. That said, again, I'm not sure how important the Ken story is; it if takes more than another hour or so, I'd just stop and do something else.

OAGr:
I'll write some more stories soon.

OAGr:
I promoted everything on master to prod, so you should be able to debug this.

RomanGalochkin:
Ok, thanks.

Issue #926, opened 2019-12-20
Add Measurable "Name", "Description", and "Owner", in the "details" tab.


OAGr:
![image.png](https://images.zenhubusercontent.com/5ab55dea4b5806bc2bcc284b/8cd3abb5-8309-4d70-88c8-b5b8e1a0ee51)

Issue #798, opened 2019-10-27
Automatically copy link to clipboard when pressing "Link to this question" [2, 2]


lagerros:
This happen when viewing a question. Would be nice!

lagerros:
Or add a button for copy-to-clipboard next to measurableId in question details.

PR #932, opened 2019-12-20
Improvements/926

RomanGalochkin:
#798 
#926


OAGr:
I realized that this may not actually complete the automatic-copying issue.
https://github.com/foretold-app/foretold/issues/798

PR #943, opened 2019-12-24
Improvements/899

RomanGalochkin:


packages/client/src/contexts/Primary.re
```diff
@@ -655,23 +667,26 @@ module CompetitorType = {
...
     };
   };
 
-  let competitorTypeInitValue =
-      (~isOwner: bool, ~state: option(Types.measurableState)) =>
-    switch (isOwner, state) {
+  let competitorTypeInitValue = (~measurable: Types.measurable) => {
+    let isAdmin = ChannelMembershipRole.isAdmin(measurable.channel);
+    let iAmOwner = measurable.iAmOwner |> E.Bool.O.toBool;
+    switch (iAmOwner || isAdmin, measurable.state) {
```
OAGr: This could be cleaned up a bit;
I suggest a function somewhere else for ``measurable.iCanEdit``

packages/server/src/graphql/middlewares/measurables.js
```diff
@@ -64,8 +67,26 @@ async function measurableNameValidation(root, args, _context, _info) {
...
 }
 
+/**
+ * @param {*} root
+ * @param {object} args
+ * @param {Schema.Context} context
+ * @param {object} _info
+ * @return {Promise<boolean>}
+ */
+async function measurableStateValidation(root, args, context, _info) {
```
OAGr: I don't like this name, and am confused by what it is validating. Is it validating that it is open to new non-comment measurements? If so, it should be renamed something like, ``measurableCanBeMeasuredValidation`` or something else specific.

packages/server/src/graphql/middlewares/measurements.js
```diff
@@ -28,23 +27,17 @@ async function measurementValueValidation(root, args, _context, _info) {
 /**
  * @param {*} root
  * @param {object} args
- * @param {Schema.Context} context
+ * @param {Schema.Context} _context
  * @param {object} _info
  * @return {Promise<boolean>}
  */
-async function measurableStateValidation(root, args, context, _info) {
+async function measurementCompetitorTypeValidation(
```
OAGr: I'd also be curious to get a better name for this and other things here as well. Is this dealing with authorization maybe? If so, can be called something to do with authorization?

OAGr:
I think most of this looks good, but I'd like to see some clarity of names.

Also, just to check, the "Edit" tab correctly shows up for admins, right? (Making sure that this story allows admins to make edits with the UI, as opposed to only doing the backend parts)

packages/client/src/contexts/Primary.re
RomanGalochkin: It was one of the hardest question in my work this year. It is not a problem to add another helpful property into measurable type. It was a question how to build a subsystem of another type of security. For instance, what exactly competitor types are available for this person.

RomanGalochkin: I programming not more than 20% of my time and all another time I am trying to think about the application.

RomanGalochkin: For instance, if I am an admin of this channel what properties I can change of the entities? Here, now we should be careful. Will think about this tomorrow more.

OAGr: Makes sense. What do you mean entities? Like, Ken-knowledge graph entities, or the different aspects of the measurables?

packages/server/src/graphql/middlewares/measurables.js
RomanGalochkin: Yes, something wrong is in this function name. I am not a native English speaker so this is why sometimes it could have a place. This function should answer the question "Is the state of measurable in a right position for adding new measurement?". This is why I like this proposal "measurableCanBeMeasuredValidation".

OAGr: Ok, sounds good

packages/server/src/graphql/middlewares/measurements.js
RomanGalochkin: No, it is not about "authorization" at all. I think it was an error to split that previous function on two function. Thanks. Will repair.

packages/client/src/contexts/Primary.re
OAGr: Makes sense, for some things it's definitely important to make good decisions, even if it takes a while.

packages/server/src/graphql/middlewares/measurements.js
OAGr: No worries, that works

RomanGalochkin: On my first glance, I thought that the function has two logic purpose this is why I split this function on two. It was an error.

packages/client/src/contexts/Primary.re
RomanGalochkin: The entities are Measurable (question), Measurement, Agent, Bot, User, Channel, Invitation, Notification and so on. So, in a word, it is a virtual objects.

RomanGalochkin:
Yes, "Edit" tab is shown for the admins and owners roles both. And the same policy on the server-side.

RomanGalochkin:
It was definitely the best question. The real question about policies. Will learn more about this later on my weekends.

RomanGalochkin:
I have fixed issues, recovered function and added another one.

OAGr:
Great, this looks good to me.

PR #944, opened 2019-12-24
Improvements/937

RomanGalochkin:
Please review the previous PR.

### DIFF
--- a/packages/client/src/components/measurables/MeasurableForm.re
+++ b/packages/client/src/components/measurables/MeasurableForm.re
@@ -557,7 +557,7 @@ module Edit = {
                   },
                   (),
                 )##variables,
-              ~refetchQueries=[|"getMeasurables"|],
+              ~refetchQueries=[|"getMeasurables", "getMeasurable"|],
               (),
             )
             |> ignore;
--- a/packages/client/src/components/measurables/MeasurableForm.re
+++ b/packages/client/src/components/measurables/MeasurableForm.re
@@ -35,7 +35,7 @@ let schema =
     Custom(
       Name,
       values =>
-        Js.String.length(values.name) > 64
+        Js.String.length(values.name) > 512
           ? ReSchema.Error("Keep it short!") : Valid,
     ),
     Custom(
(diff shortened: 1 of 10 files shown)


packages/client/src/components/measurables/MeasurableForm.re
```diff
@@ -557,7 +557,7 @@ module Edit = {
                   },
                   (),
                 )##variables,
-              ~refetchQueries=[|"getMeasurables"|],
+              ~refetchQueries=[|"getMeasurables", "getMeasurable"|],
```
OAGr: good catch

packages/server/migrations/js/201912230755000-notebooks-index.js
```diff
@@ -32,6 +35,10 @@ module.exports = {
         'Notebooks_name_ownerId_unique',
       );
 
+      // @todo: Here first of all we should have to remove
+      // @todo: duplicates and only then to create another
+      // @todo: unique index.
```
OAGr: When do you expect that we do these @todos? Can you make a new story or two and prioritize accordingly?

packages/server/src/graphql/types/measurables.js
```diff
@@ -7,15 +7,37 @@ const resolvers = require('../resolvers');
...
+const name = {
+  type: graphql.GraphQLNonNull(scalars.string3to512),
+};
+const labelSubject = {
+  type: graphql.GraphQLString,
+  description: 'Description of a question\'s entity',
+};
+const labelOnDate = {
+  type: DateType.default,
+  description: 'Description of a question\'s date',
```
OAGr: Kudos for adding these descriptions!

packages/server/migrations/js/201912230755000-notebooks-index.js
RomanGalochkin: It is a todo not in a direct sense. It is like a remark. I am not sure that we should do something here this is why I did not create a new issue.

OAGr: Ok, that works

PR #945, opened 2019-12-25
Reduces a notebook sidebar space

RomanGalochkin:
Please review previous pull requests.

### DIFF
--- a/packages/client/src/components/SLayout.re
+++ b/packages/client/src/components/SLayout.re
@@ -1,5 +1,7 @@
 open Utils;
 
+type container = [ | `fixedWidth | `fluid | `none];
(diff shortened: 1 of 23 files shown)


OAGr:
Looks like this failed on Travis
image

RomanGalochkin:
Will investigate.

packages/client/src/components/SLayout.re
```diff
@@ -1,5 +1,7 @@
 open Utils;
 
+type container = [ | `fixedWidth | `fluid | `none | `fluidLeft];
```
OAGr: I like this as the solution, good idea

PR #946, opened 2019-12-25
Hides columns

RomanGalochkin:


OAGr:
lgtm

RomanGalochkin:
Will close temporary.

RomanGalochkin:
Screenshot_20191226_093854

And may it be it has a sense to show a distribution like this.

OAGr:
Good point, though we'd want to keep the width fixed then.

PR #949, opened 2019-12-27
URL

RomanGalochkin:


packages/client/src/pages/measurables/MeasurableIndex__Logic.re
```diff
@@ -6,7 +6,30 @@ module ReducerConfig = {
...
-  let getId = (params: itemType) => params.id;
+  let getId = (itemType: itemType) => itemType.id;
+
+  let onItemDeselected = (params: callFnParams) => {
+    let st =
+      params.states
+      |> MeasurableQueryIndex.make
+      |> MeasurableQueryIndex.toUrlParams;
+    let channelLink = Routing.Url.toString(ChannelShow(params.channelId));
+    ReasonReact.Router.replace(channelLink ++ st ++ "#!");
```
OAGr: What's the point of the "#!"?

OAGr:
This changes the URL, which is good.

It doesn't seem to handle history that well. When I press "back" in my history, it takes me to the community screen, instead of to the previous question. When I then press "forward", it kind of messes up.

packages/client/src/pages/statics/NotFoundPage.re
```diff
@@ -1,6 +1,10 @@
 [@react.component]
 let make = () => {
   <SLayout container=`fluid>
-    <Center> {"Sorry, but the page is not found :(" |> Utils.ste} </Center>
+    <Center>
+      <Icon icon="PACMAN" size="2em" />
```
OAGr: Why pacman?

OAGr:
Can you explain how this works a bit?

OAGr:
I'm going to merge this because I'm pretty sure it's an improvement. But I could see improving it further, as stated, so please do respond to the comments when you next get a chance.

packages/client/src/pages/statics/NotFoundPage.re
RomanGalochkin: I do not know, just like Pacman. I want to give to the application more interconnection with users through messages and dialogues. To make it more interactive.

RomanGalochkin:
The main issue was to divide the events about URL changing from simple replacing URL in browser's history. Well, there is no way that I know how to do this at least. 

1. We update URL and add "#!" at the end of URL.
2. We update view components to show something.
3. We get event about URL changes and see at the end of the URL and do nothing if the end equals to "#!".

It is a temporary solution. I want to change "Paginator" step by step to make it more widely used and want to study it to recover the state from the URL in the future when it will be necessary.

packages/client/src/pages/measurables/MeasurableIndex__Logic.re
RomanGalochkin: To make it possible to differentiate it between two types of events. Which are when you push something to browser's history and another one when you replace URL.

RomanGalochkin:
About back/forward buttons. Yes, in the next stories we could change it. May be to write an adapter from Reason to JS to handle all browser's events including "push events" and "replace events". The good idea did not come it before.

PR #971, opened 2020-01-10
Improvements/956

RomanGalochkin:
There were no huge issues except only two moments. 1. There were really 256 questions with "expected resolution date" of the 30th of December. 2. The application was not ready to run in multi-instance version.


OAGr:
Ok, will merge after Travis passes

PR #979, opened 2020-01-15
Feature/959 3

RomanGalochkin:
I have done it with direct deleting from DB but I do not like it. The best way is to mark "notebook" as "deletedAt".


packages/client/src/contexts/Lang.re
```diff
@@ -15,4 +15,5 @@ let measurableIsSaved = "Question is saved.";
 let copiedMessage = "Copied";
 let copiedDescription = "The question ID was copied.";
 let backButtonTip = "Back";
-let pageIsNotFound = "Sorry, but the page is not found.";
\ No newline at end of file
+let pageIsNotFound = "Sorry, but the page is not found.";
+let notebookRemove = "Are you sure delete this notebook?";
```
OAGr: Change to "Are you sure you want to delete this notebook?"

packages/client/src/contexts/Lang.re
```diff
@@ -15,4 +15,5 @@ let measurableIsSaved = "Question is saved.";
 let copiedMessage = "Copied";
 let copiedDescription = "The question ID was copied.";
 let backButtonTip = "Back";
-let pageIsNotFound = "Sorry, but the page is not found.";
\ No newline at end of file
+let pageIsNotFound = "Sorry, but the page is not found.";
```
OAGr: Change to "Page not found"

packages/client/src/contexts/Lang.re
RomanGalochkin: Thanks.

OAGr:
Great, will merge after it passes Travis.

OAGr:
> I have done it with direct deleting from DB but I do not like it. The best way is to mark "notebook" as "deletedAt".

I semi-agree. I think there are some nicities of actually deleting the data; for the sake of allowing the user the ability to make sure that we don't store it anywhere. 

In the future I'd be find changing this to "archive".

That said, for Notebooks, I'm not too worried, because it's easy enough just to copy the main markdown content.

PR #980, opened 2020-01-16
Flex d3

RomanGalochkin:
It is weird but when I used "react-use" the hooks did not work. So I just used the browser events to catch resize moments. I am still not sure about something and going to add a few changes.


OAGr:
Gr... maybe one of the other links would work?

If we could have nice ways to wrap hooks, that seems useful going forward, as some of the other hooks could later be neat to add.

RomanGalochkin:
I think this is because I do not understand properly something. Will think more. The errors were like in previous problems with hooks when Reac says that hooks are broken.

RomanGalochkin:
It turned out that even we use the same version of the React the parcel will duplicate it inside the bundle file while we do not use "aliases" in the main application. So previous battle with "hooks are wrong" maybe could resolve the same way.

packages/client/package.json
```diff
@@ -80,14 +80,18 @@
...
+    "react-dom": "16.12.0",
     "reschema": "^1.3.0",
     "reason-react": "git+https://github.com/RomanGalochkin/reason-react.git#0.7.0.beta.0",
     "parcel-plugin-bundle-visualiser": "1.2.0",
     "parcel-plugin-html-externals": "0.1.2"
   },
+  "alias": {
+    "react": "./node_modules/react",
+    "react-dom": "./node_modules/react-dom"
+  },
```
OAGr: Can you explain the use of these aliases?

packages/client/src/components/forms/CdfInput.re
```diff
@@ -487,10 +487,10 @@ module Main = {
       <div className=Styles.chartSection>
         {switch (state.dataType, state.cdfType) {
          | ("FLOAT_CDF", _) when E.A.length(state.floatCdf.xs) > 1 =>
-           <FC__CdfChart__Large cdf={state.floatCdf} />
+           <FC__CdfChart__Large cdf={state.floatCdf} width=Some(0) />
```
OAGr: Why is the width 0? That seems odd to me.

packages/components/package.json
```diff
@@ -21,17 +21,20 @@
     "license": "MIT",
     "dependencies": {
```
OAGr: Why do we need this file at all? I would have hoped we could just have package.json files in each subcomponent's files.

packages/components/src/Charts/cdfChart.js
```diff
@@ -19,48 +21,45 @@ function getRandomInt(min, max) {
...
-  componentDidMount() {
-    this.drawChart();
-  }
-
-  componentDidUpdate() {
-    this.drawChart();
-  }
+function CdfChart(props) {
+  const id = "chart-" + getRandomInt(0, 100000);
+  const [sized, { width }] = useSize(() => {
```
OAGr: kudos for getting this work!

OAGr:
Merged to use this, but please do respond to comments sometime.

packages/client/package.json
RomanGalochkin: The Parcel uses this configuration to resolve the paths of libraries to a single place.

packages/client/src/components/forms/CdfInput.re
RomanGalochkin: It was done to save backward compatibility. Later we could refactor this a bit more.

PR #992, opened 2020-01-21
Removes Jsx2 from components

RomanGalochkin:
I had better to use global imports in some cases to reduce the number of using module prefix "ForetoldComponent" but will do it later on-demand.


OAGr:
Happy to see this.

PR #993, opened 2020-01-22
Titles

RomanGalochkin:
In some cases, we cannot use nested React Hooks this is why here I added new component as React Function component. And plus I tried to use context with hooks and rewrote it so now this is in the code though it is not necessary in this PR.


packages/client/package.json
```diff
@@ -82,8 +82,8 @@
     "react-copy-to-clipboard": "5.0.2",
     "react": "16.12.0",
     "react-dom": "16.12.0",
-    "reschema": "^1.3.0",
-    "reason-react": "git+https://github.com/RomanGalochkin/reason-react.git#0.7.0.beta.0",
+    "reschema": "1.3.0",
+    "reason-react": "0.7.0",
```
OAGr: Kudos for getting us back on the main branch

packages/client/src/pages/Navigator.re
```diff
@@ -1,75 +1,76 @@
...
 
-        // Channels
-        | (Channel(channel), _) => <ChannelNavigation channelPage=channel />
-        | (ChannelIndex, _) =>
-          <FillWithSidebar> <ChannelIndex /> </FillWithSidebar>
-        | (ChannelNew, Some(_)) =>
-          <FillWithSidebar> <ChannelNew /> </FillWithSidebar>
+  <>
+    {switch (route, context.loggedUser) {
+     // Statis Pages
```
OAGr: *Static* I think?

RomanGalochkin: Yes, sure.

packages/client/src/pages/Title.re
```diff
@@ -0,0 +1,42 @@
...
+    ChannelGet.component2(~id=channelId, result =>
+      switch (result) {
+      | Success(channel) =>
+        let title = [|channel.name, Lang.Title.main|] |> E.Title.toString;
+        <ForetoldComponents.Title title />;
+      | _ => <Null />
+      }
+    )
+  | ChannelIndex =>
+    <ForetoldComponents.Title
```
OAGr: I think some more dryness / functional additions would be nice.

Maybe one function that returns an Option(string) with the title, and another that wraps that with <ForetoldComponents.Title/>

RomanGalochkin: No, we should use Factories this way. Only components.

RomanGalochkin: I have done because of loading channel data with components. I do not want to divide "switch" construction on two and do it in a complicated way. It is not readable.

RomanGalochkin: Although it should be cleaned. Will think about it later.

OAGr: If nothing is quickly obvious, don't worry about it.

RomanGalochkin: Yes, please let us leave this for now later I will fix this when I find some useful pattern.

PR #1001, opened 2020-01-24
The Votes (server-side)

RomanGalochkin:


packages/server/migrations/js/202001221523000-unlogged.js
```diff
@@ -0,0 +1,45 @@
...
+    try {
+      await queryInterface.sequelize.query('BEGIN');
+
+      const tablesWithOrder = [
+        'Votes', 'Mutexes', 'FeedItems',
+        'NotificationStatuses', 'Notifications',
+      ];
+      for (const table of tablesWithOrder) {
+        await queryInterface.sequelize.query(
+          `ALTER TABLE "${table}" SET UNLOGGED;`,
```
OAGr: What does this do?

RomanGalochkin: This optimizes tables inside of Postresql. It will work a bit faster. "UNLOGGED" means that Postresql will not write logs for this table in the recovery log.

packages/server/src/data/votes-data.js
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /**
+   * @param {Models.AgentID} agentId
+   * @param {Models.MeasurementID} measurementId
+   * @returns {Promise<Models.Vote>}
+   * @public
+   */
+  async downvote(agentId, measurementId) {
+    return this._changeVote(agentId, measurementId, this._down);
```
OAGr: From what I can tell, the client is expected to send ``upvote`` and ``downvote`` requests. Each one increments or decrements the votes by 1. 

I think this makes me a bit uncomfortable. I could imagine that voting will lag a bit; so when a user clicks "Upvote" 3 times, it will take a while for each to reach the backend.

What would you think of this approach:
The client instead sends a total vote count, like "-4" or "+4". This is sent to the server when they stop clicking (there's a short debouncing).

RomanGalochkin: Ok, will fix it.

packages/server/src/data/votes-data.js
```diff
@@ -0,0 +1,163 @@
...
+   * @param {Models.AgentID} agentId
+   * @param {Models.MeasurementID} measurementId
+   * @param {Layers.Transaction} transaction
+   * @returns {Promise<*>}
+   * @protected
+   */
+  async _vote(agentId, measurementId, transaction) {
+    const params = new Params({ agentId, measurementId });
+    const query = new Query();
+    const options = new Options({ transaction, raw: true });
```
OAGr: If the vote is 0, do you think we should delete the vote? A vote of 0 should act identical to the lack of any vote.

RomanGalochkin: Since we just sum all votes it is not necessary to remove the vote.

OAGr: Sure, I guess the cost should be low.

OAGr:
Also, note that this may have failed Travis.

RomanGalochkin:
Yes, sure. Sometime Travis falls.

RomanGalochkin:
I have changed the interface. Now it looks like:
```
mutation($measurementId: MeasurementId!, $input: VoteInput!) {
  vote(measurementId: $measurementId, input: $input) {
    id
    voteAmount
  }
}
```
`{"measurementId":"73bd5b3a-0b2c-4e51-9790-a783ec1241f3", "input": {"voteAmount": 3}}`

OAGr:
Great, that works

PR #1020, opened 2020-01-30
Adds a prediction column for the notebook side bar

RomanGalochkin:
Screenshot_20200130_124522


OAGr:
I think this isn't quite right. This would re-introduce the "10 to 30" notation when the questions are continuous, which is not intentional.

RomanGalochkin:
I have mentioned this PR in the slack.

PR #1022, opened 2020-01-30
Improve debouncer for nicer voting

OAGr:
We don't want users to have to wait for each vote for it to ping the server. This PR addresses that by separating the state of the vote widget from the backend server calls.

### DIFF
--- a/packages/client/src/components/measurements/AgentVote.re
+++ b/packages/client/src/components/measurements/AgentVote.re
@@ -30,11 +30,13 @@ module MeasurementVotes = {
       });
 
     <>
-      <Antd.Button onClick={_ => modify(Down)} disabled={!canDecrement}>
+      <Antd.Button
+        size=`small onClick={_ => modify(Down)} disabled={!canDecrement}>
         {"<" |> Utils.ste}
       </Antd.Button>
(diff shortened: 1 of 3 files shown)


packages/client/src/components/measurements/AgentVote.re
```diff
@@ -0,0 +1,76 @@
...
+      <Antd.Button
+        size=`small onClick={_ => modify(Up)} disabled={!canIncrement}>
+        {">" |> Utils.ste}
+      </Antd.Button>
+    </>;
+  };
+};
+
+// TODO: I think this would be cleaner if this module were a wrapper, similar to the <Mutation> used inside.
+// I wasn't sure how to do this though.
```
RomanGalochkin: I guess we could try to use Reason Apollo Hooks. Maybe it would help us to clean it up.

RomanGalochkin:
Looks good to me.

packages/client/src/components/measurements/AgentVote.re
OAGr: Good point. Do you think hooks could work for most of our mutation logic? The wrapping components seem like a messy method.

PR #1024, opened 2020-01-31
A votes popover

RomanGalochkin:


packages/client/src/components/measurements/MeasurementsTable.re
```diff
@@ -120,13 +120,87 @@ module Helpers = {
...
+        ]);
+      let subRow = style([width(`percent(100.))]);
+    };
+    module FC = ForetoldComponents;
+    [@react.component]
+    let make = (~measurement: Types.measurement) => {
+      MeasurementVotesGet.component(
+        ~measurementId=?Some(measurement.id), measurementVotes =>
+        measurementVotes
+        |> Array.mapi((index, measurementVote: Types.vote) => {
```
OAGr: I think I'd prefer we use E.Array instead of Array. I don't trust the standard library that much.

OAGr:
This looks quite good to me, nice work!

packages/client/src/components/measurements/MeasurementsTable.re
RomanGalochkin: Me too. Fixed.

PR #1037, opened 2020-02-04
Total Votes

RomanGalochkin:
This PR should make the application works faster than previous versions.


packages/client/graphql_schema.json
```diff
@@ -5080,6 +5080,18 @@
               "isDeprecated": false,
               "deprecationReason": null
             },
+            {
+              "name": "totalVotes",
```
OAGr: Can we rename this to "totalVotesReceived", instead of "totalVotes", to make sure that it's more evident what this refers to? Later we may want an attribute for "totalVotesGiven", for reasons mentioned earlier in the story comments.

OAGr:
The interface looks mostly good, one minor comment.

packages/client/graphql_schema.json
RomanGalochkin: Sure, thanks.

RomanGalochkin:
Updated.

OAGr:
It's breaking travis, seems related.

image

RomanGalochkin:
Yes, fixing.

RomanGalochkin:
Fixed.

PR #1061, opened 2020-02-11
Feature/1038

RomanGalochkin:


OAGr:
Looks good to me. Note that Notebook bookmarks are coming up, so maybe there will be some db refactoring there (If you want to use one table for both, but I could go either way)

PR #1089, opened 2020-02-25
Display some statistics below the distribution chart

NunoSempere:
It has the expected value, median, confidence intervals & the point of highest likelihood.

Missing:
- Comparisons to the most recent aggregate
- Tests
- Ability to deal gracefully with very wide distributions.


packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+            <tr>
+                <td>{React.string("Highest likelihood")}</td>
+                <td>{Display.highestLikelihood(cdf)}</td> 
+                //<td>{React.string("Aggregation Value")}</td>
+            </tr>        
+            
+        </tbody>
+    </table>;
+
+    /* <p>{React.string("The lone and level sands stretch far away")}</p>; */
```
OAGr: Please remove comments before submitting.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+                | someOtherIdex => element -. Array.get(ys,index-1);
+            }
+
+            }, ys); 
+            
+            ysPdf;
+        } 
+        let expectedValue = (cdf: ForetoldComponents.Types.Dist.t) => {
+            let ys= toPdf(cdf);
+            let result = Js.Array.reducei((accumulator, x, index) => (accumulator +. (x *. Array.get(ys,index)) ), 0.0, cdf.xs)
```
OAGr: I think this makes the assumption that samples are distributed across the x-axis evenly, which may not always be the case.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
+[@react.component]  
+let make =  
+    (
+        ~cdf: ForetoldComponents.Types.Dist.t, 
+        ~measurable: Types.measurable
+    ) => {
+    
+    module Calculate = {
```
OAGr: The formatting of this document is kind of weird. I recommend using something like reason-vscode to auto-format on save.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
+[@react.component]  
+let make =  
+    (
+        ~cdf: ForetoldComponents.Types.Dist.t,
```
OAGr: In general, I'd recommend against using the types from ForetoldComponents in this app, unless we're using them in order to interact with these components directly.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+            pos==(-1) ? None : Some(cdf.xs[pos]);
+        }
+
+        let confidenceInterval = (cdf: ForetoldComponents.Types.Dist.t,p) => {
+            let ys = cdf.ys;
+            let low_p = ((100.0-.p)/.100.0)/.2.0;
+            let high_p = 1.0-.low_p;
+            
+            let low_value_index = Js.Array.findIndex( y => y>=low_p, ys);
+            let high_value_index = Js.Array.findIndex( y => y>=high_p, ys);
```
OAGr: Why did you use Js.Array and not something else? For most functions like these, I'd prefer we use something within E.A, where we store the main Array functions (and if something important isn't there yet, then feel free to add it there)

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+            let ys= toPdf(cdf);
+            let result = Js.Array.reducei((accumulator, x, index) => (accumulator +. (x *. Array.get(ys,index)) ), 0.0, cdf.xs)
+            result;
+        } 
+
+        let median = (cdf: ForetoldComponents.Types.Dist.t) => {
+            let ys = cdf.ys;
+            let pos = Js.Array.findIndex( y => y>=0.5, ys);
+            pos==(-1) ? None : Some(cdf.xs[pos]);
+        }
```
OAGr: I know we already have code to find specific confidence intervals.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+            Calculate.highestLikelihood(cdf)
+            -> genericDisplay;
+        }
+
+    };
+    
+    <table style={Display.tableCss}>
+        <thead>
+            <tr>
+                <th>{React.string("")}</th>
```
OAGr: Do you need an empty string here? Can it just be empty inside?

packages/client/src/components/measurements/MeasurementCdfInput.re
```diff
@@ -460,10 +460,28 @@ module Main = {
...
+                  width={Some(0)}
+                  />
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  <MeasurementStatisticsBelowChart
+                  cdf={state.floatCdf}
+                  measurable={measurable}
+                  />
```
OAGr: Why are these items in a table? Do you just need one on top of the other?

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+
+            let low_value = Array.get(cdf.xs,low_value_index);
+            let high_value = Array.get(cdf.xs,high_value_index);
+
+            (low_value, high_value);
+        }      
+
+        let highestLikelihood = (cdf: ForetoldComponents.Types.Dist.t) => {
+            let ysPdf = toPdf(cdf) 
+            let indexOfMaxValue = Js.Array.reducei((iMax, x, i) => (x > Array.get(ysPdf,iMax))? i : iMax, 0, ysPdf);
```
OAGr: I think we can expect the ys and xs to be in order, so if you want the highest, just get the last.

packages/client/src/components/measurements/MeasurementStatisticsBelowChart.re
```diff
@@ -0,0 +1,131 @@
...
+            let ys =  cdf.ys;
+            let ysPdf = Array.mapi((index, element) => {
+            switch(index){
+                | 0 => element;
+                | someOtherIdex => element -. Array.get(ys,index-1);
+            }
+
+            }, ys); 
+            
+            ysPdf;
```
OAGr: I think the Distribution.T.toPdf function does this.

PR #1132, opened 2020-04-07
Feature/1127

RomanGalochkin:
It is temporarily hard to develop the client-side. What is annoying. For the first step, we should run "yarn pb" from the project root and then we are ready to run client commands "yarn bsb-watch", "yarn parcel". However, at least I could upgrade "bs-platform", now we are sure that this does not correspond to "duplicated packages" issue.


OAGr:
Thanks. Yea, the client-side has become really frustrating. Sorry to hear that upgrading bs-platform does not fix that duplicated package issue.

PR #1135, opened 2020-04-09
Formatted packages/cdf/dist/index.js with an autoformatter

NunoSempere:
File was terribly formatted before, and in fact crashed Github/Visual Studio code when one tried to open it. The autoformatter used was: https://www.npmjs.com/package/js-beautify


OAGr:
Thanks for the suggestion!

I believe this file is autogenerated (from running ``parcel build``), and meant to be minified/compressed. The files you are supposed to read are in /lib/, not /dist/. If we continue working on this, all new versions will also be compressed, so this would be a very temporary change, and I'm not sure we want formatted minified code (I think it does that for a reason)

PR #1137, opened 2020-04-09
Epic: Verified Channels

RomanGalochkin:


packages/server/src/data/models/model-postgres.js
```diff
@@ -829,6 +845,85 @@ class ModelPostgres extends Model {
...
+   * @return {string}
+   */
+  _verifiedMeasurements(name = '') {
+    return `(
+      /* Verified Channel Memberships (${name}) */
+      SELECT "ChannelMemberships"."agentId"
+      FROM "ChannelMemberships"
+      LEFT JOIN "Channels" ON "Channels"."id" = "ChannelMemberships"."channelId"
+      WHERE "ChannelMemberships"."isVerified" = TRUE
+      OR "Channels"."requireVerification" != TRUE
```
OAGr, line 897: Why are you using direct SQL instead of sequelize here?

packages/server/src/enums.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+/**
+ * @todo: He is a small issue. If we have "viewers"
+ * @todo: so then why then can add predictions?
+ * @todo: The better to give them another
+ * @todo: group name.
+ *
+ * @todo: But here is other thing. If we have
+ * @todo: "requireVerification" then this is ok.
```
OAGr, line 28: The word "VIEWER" no longer makes sense (that was for an older version). Maybe "UNCERTIFIED_FORECASTER" and "CERTIFIED_FORECASTER"?

packages/server/src/data/models/model-postgres.js
RomanGalochkin: Because Sequelize has its own edges of usage. And for this purpose, this is not a good library. Sequelize is a good library for small and simple application. In the future, I would recommend to remove Sequelize at all either to reduce its usage.

OAGr: Makes sense. Are there other libraries you'd recommend instead, or just use raw SQL much more?

RomanGalochkin: It depends on issues. Do not think about SQL / Sequelize much. SQL is metalanguage between application and database. First of all thing about application architecture: data layers (restrictions, filters), middleware, validators, authorizes, graphql interface, resolvers, data definitions, sync flows.

I do not imagine how to explain this better.  So for this step of application growing the Sequelize is ok. I think the most use cases are coded in the code-base. Now Sequelize is a more DB driver and is used under Data / Models layers.

packages/server/src/enums.js
RomanGalochkin: Mmm, I am not sure and this is why. If we have a channel setting "requireVerification" we should save "viewer" group. Either we should remove the new channel setting and add two groups "UNCERTIFIED_FORECASTER" and "CERTIFIED_FORECASTER". But what if we will add another option for the channel? Then according to our logic, we should add new groups and add groups to combine previous logic so we would have 6 groups or more.

OAGr: Sorry, I think I misunderstood the comment before.

PR #1143, opened 2020-04-13
Fixes types

RomanGalochkin:
Besides, I have removed measurements (33) with nulls right in the DB. These measurements were created at the same time.


OAGr:
Thanks!

PR #1145, opened 2020-04-14
Order leaderboard

NunoSempere:
Orders the leaderboards by Total Net Log Score.

order_leaderboard


OAGr:
Nice, looks good.

One issue is that ideally this would probably be done on the server side instead; if there are more than 20 items, then I assume ordering won't hold between pages.

Happy to delay on that though.

NunoSempere:
Just checked, it indeed doesn't preserve the ordering between pages.

OAGr:
Would you want me to merge it without the backend change? It seems nice for a few items, but possible confusing. In general I don't like confusing.

RomanGalochkin:
So why not, I think if it will be needed we implement this on the server-side.

OAGr:
@RomanGalochkin how difficult would this be to do on the server side?

RomanGalochkin:
Give me a minute.

RomanGalochkin:
At this moment this feature will require some hours (more than 4) since this property is not stored in DB.

OAGr:
Good to know. I don't think it's worth that much now, maybe later though.

OAGr:
I'm going to close this, can reinspect when server change is implemented.

NunoSempere:
Nothing against.

PR #1172, opened 2020-04-28
Markdown preprocessing

NunoSempere:
Jacob Laguerros said on slack:
> I think that if one writes single lines between sentences, they get collapsed to a single text block.
> So if I type:

```
My first point.
My second point.

My third point.
```

> This gets rendered as:
```
My first point. My second point.
My third point.
```
> This makes it harder to read long messages [...] Could we change it?

This pull request adds newlines as the comments are sent to the database so that this is not a problem anymore, for new messages / comments.


packages/client/src/components/measurements/MeasurementCdfInput.re
```diff
@@ -466,6 +466,15 @@ module Main = {
   };
 };
 
+module MarkdownPreprocessing = {
+
+  let addNewLines = description: string => {
+    let duplicateNewLines = Js.String.replaceByRe([%re "/[\\n]/g"], "\n\n", description);
+    let pruneCuatrupleLines = Js.String.replaceByRe([%re "/[\\n\\n\\n\\n]/g"], "\n\n", duplicateNewLines);
```
OAGr, line 473: "quatruple" -> "quadruple". 
Also, This would make it impossible for anyone who wanted 4 newlines, I assume?

OAGr: This seems like a decently hacky situation to me. I realize it's a tricky thing though.

packages/client/src/components/measurements/MeasurementForm.re
```diff
@@ -16,7 +16,7 @@ let make = (~measurable: Types.measurable, ~defaultValueText="") => {
               measurable.id,
               value,
               competitorType,
-              description,
+              MeasurementCdfInput.MarkdownPreprocessing.addNewLines(description),
```
OAGr, line 19: If we do this, we should probably have it apply to all markdown renders, not just measurement inputs.

packages/client/src/components/measurements/MeasurementCdfInput.re
NunoSempere: You can't have 4 newlines right now. Try it.

packages/client/src/components/measurements/MeasurementForm.re
NunoSempere: > I think you've convinced me that this is a justified feature, but I think doing it well and right will be more work.

Plausibly. Another possibility would be to move to Github flavored markdown, or to some implementation (https://simplemde.com/?) which doesn't have the line break problem.

NunoSempere: Agree that it makes sense to have it on other renderers too

OAGr: I think Github does do this for new lines. I would prefer to move to Github flavored or another standard one that has them, instead of us coming up with our own hybrid, but was assuming that would be more work.

packages/client/src/components/measurements/MeasurementCdfInput.re
NunoSempere: (Fixed the "quatruple" -> "quadruple" typo)

NunoSempere: Does the solution seem less hacky to you given that the 4 newlines is not a problem?

Some other things I was worried about were:
- Ordered lists breaking if you added an additional newline
- Tables

But I checked for that (https://www.foretold.io/c/e351f8f3-cf9c-4787-a787-68257220084d/m/012b73bf-82d2-4c70-9c85-644f2685fc68), and it is not a problem.

packages/client/src/components/measurements/MeasurementForm.re
NunoSempere: So coming up with our hybrid might be a nightmare. For example, if you want to preserve newlines within codeblocks (\`\`\`), then you either have to write a parser, or some really ugly regular expressions. I was briefly looking into the Github Flavored Markdown specs: https://github.github.com/gfm, and I'm just glad someone else did it.

In this case I'd push for either.
1. Just do the hack for comments, but maybe not for notebooks
2. Move to github flavored markdown, or to another specification and/or a better editor.

I have some intuitions that in this case, the hack is better, but I also wouldn't mind the second option. I dislike the option in between, the "write your own principled hybrid".

OAGr: Apparently we may be able to implement github flavored markedown with our current react-markdown package, though of course I'm not sure exactly how hard it would be. https://github.com/rexxars/react-markdown/blob/c63dccb8185869cfc73c257d098a123ef7a7cd33/demo/src/demo.js

I'm hesitant to go forward with this as in (1), it just makes me feel uneasy. Hacks like that can really bite, we've probably gone too far that type of direction already.

OAGr: It's possible we can use a plugin for this. https://github.com/rexxars/react-markdown/issues/105

NunoSempere: Oh, cool

PR #1176, opened 2020-04-29
Feature/1160

RomanGalochkin:
Screenshot_20200429_164630

All packages had been updated except these four on the screen:
- "antd" requires the upgrade of the "bs-ant-design-alt" (since this is an adapter),
- "mathjs" brokes production build somehow,
- "reason-apollo-hooks" requires much more time to rewrite some parts of code to use updated interfaces however this is possible to do,
- "bs-css" is not obvious how to update at the moment since there appear some weird parcel errors.


OAGr:
Sounds fine. I think we may want to remove antd at some point, as it's quite huge. How much work do you think it would be to update reason-apollo-hooks? Do you think that that one would be important?

RomanGalochkin:
I think it will take about 4 or fewer hours to update the client code. I think yes this upgrade of hooks we should do anyway.

PR #1203, opened 2020-05-15
Upgrades bs-platform

RomanGalochkin:


OAGr:
Great, that was fast.

PR #1207, opened 2020-05-18
Feed items

RomanGalochkin:


packages/client/src/contexts/FeedItemBody.re
```diff
@@ -9,6 +9,18 @@ type measurable = {
   measurableId: string,
 };
 
+type measurableB = {
```
OAGr, line 12: why is this called ``measurableB``?

OAGr: maybe something like ``measurableWithEntities`` if that's the main distinction?

RomanGalochkin: I knew that you will ask this question. The main reason is to differentiate the "previous measurable feed items" and "new measurable feed items". Until both versions are stored in the DB at the current moment we have to keep the both JS classes in the code. These types are similar to versions. But the better choice to use the "typification": the previous measurable is type "A", the new one is type "B". It would not be right to call new ones as "measurableWithEntities" since in the third possible edition we will need to call the third edition somehow.

RomanGalochkin: The feed items classes should be isomorphic DB data.

RomanGalochkin: Hm, after some thougths I am not sure. So will change the name to one you have proposed.

RomanGalochkin: I have updated the new measurable type. Not it is called "measurableWithEntities".

OAGr: Great, thanks!

PR #1213, opened 2020-05-25
Fixes back browser button

RomanGalochkin:
It turned out that the application has many issues with application state. So this is the first step to reduce these issues. Firstly we should make some parts of the code more clear and simpler.


OAGr:
Good point. Seems kind of gnarly. That was a lot of steps.

